PR #120, opened 2024-09-09
fix: correct conda package name

jflatt-gia:
Blender cloud submitter was defaulting with a conda package name that is inconsistent with the github and pypi package name.
 
Rename the package variables from "blender-openjd" to "deadline-cloud-for-blender".

Changes the default value.  Can be changed via deadline Conda Queue Environment, or manually in the submitter dialog.
When overriding python wheels, it should exclude the correct package.

In Blender, Render->Submit to AWS Deadline Cloud, Conda Packages line editor shows the package name.

Not applicable.

No

----

### DIFF
--- a/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
+++ b/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
@@ -76,7 +76,7 @@ def create_deadline_dialog(parent=None) -> SubmitJobToDeadlineDialog:
     adaptor_version = ".".join(str(v) for v in adaptor_version_tuple[:2])
     # Need Blender and the Blender OpenJD application interface adaptor
     rez_packages = f"blender-{blender_version} deadline_cloud_for_blender"
-    conda_packages = f"blender={blender_version}.* blender-openjd={adaptor_version}.*"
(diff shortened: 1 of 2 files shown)


epmog:
Hey jflatt-gia, thanks for creating a pull request!

Since the conda package in the `deadline-cloud` channel _is_ called `openjd-blender` this would break the default workflow in service-managed fleets since `deadline-cloud-for-blender` does not exist. Can you explain a little more about your use-case for this change? The override adaptors should also still work with the previous naming.

Is it specifically just the inconsistent naming between repo/package (deadline-cloud-for-blender) vs. adaptor application (blender-openjd) being used?

jflatt-gia:
> Hey jflatt-gia, thanks for creating a pull request!
> 
> Since the conda package in the `deadline-cloud` channel _is_ called `openjd-blender` this would break the default workflow in service-managed fleets since `deadline-cloud-for-blender` does not exist. Can you explain a little more about your use-case for this change? The override adaptors should also still work with the previous naming.
> 
> Is it specifically just the inconsistent naming between repo/package (deadline-cloud-for-blender) vs. adaptor application (blender-openjd) being used?

Okay, that makes sense, I was never able to get into the deadline-cloud channel.  Instead, I created a custom conda channel using the deadline-cloud-for-blender package name.

There's another issue where my values are being overwritten with these defaults every time.  If I can have it respect my Conda queue environment values, it would eliminate this issue.  It currently works great for the CondaChannels variable, but the CondaPackages variable gets replaced in shared_job_settings_tab.py using these default values.

PR #122, opened 2024-09-19
ci:package and publish Blender plugin to the GitHub release

moorec-aws:
1. [__init__.py](https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/__init__.py#L24) contains the version of the latest release. Currently this has to be updated manually.
2. Users are required to git clone the release branch to use the latest submitter.

1. Updated __init__.py with the release version when the bump workflow is run.
2. Zip the contenets of `addons` into the dist_extras directory during the release process.

1. Proper versions updated in plugin file when we do a release
2. Users can access the plugin from the release page.

Tested in a development environment using the [reusable_bump](https://github.com/aws-deadline/.github/blob/mainline/.github/workflows/reusable_bump.yml) and [reusable publish](https://github.com/aws-deadline/.github/blob/mainline/.github/workflows/reusable_publish.yml) workflows.

* Confirmed that the versions are updated properly in the ChangeLog PR.
* Confirmed that the zip exist on the release page with the expected contents.

No

No

----

### DIFF
--- a/.github/actions/prepush_release_hook/action.yml
+++ b/.github/actions/prepush_release_hook/action.yml
@@ -1,0 +1,18 @@
+inputs:
+  semver:
+    required: true
+    type: string
+  tag:
+    required: true
+    type: string
+
+runs:
+  using: composite
+  steps:
+  - name: Zip submitter plugin
+    shell: bash
+    run: |
+      mkdir dist_extras
+      cd src/deadline/blender_submitter/addons
+      zip -r  ../../../../dist_extras/blender_submitter_plugin_${{inputs.semver}}.zip .
+      cd ../../../../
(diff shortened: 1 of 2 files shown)


AWS-Samuel (approved):

.github/actions/prepush_release_hook/action.yml
```diff
@@ -0,0 +1,18 @@
...
+runs:
+  using: composite
+  steps:
+  - name: Zip submitter plugin
+    shell: bash
+    run: |
+      mkdir dist_extras
+      cd src/deadline/blender_submitter/addons
+      zip -r  ../../../../dist_extras/blender_submitter_plugin_${{inputs.semver}}.zip .
+      cd ../../../../
```
epmog, line 18: nit: can use pushd/popd here, though I don't think you need to do the last cd when it's encapsulated in the bash script. Doesn't actually affect anything here though.

epmog (approved):

Issue #105, opened 2024-05-05
Bug: Directory auto-detected as file; submission fails


michaeldavie:
### Expected Behaviour

Submission succeeds

### Current Behaviour

Job submission fails because the fluid simulation cache is listed as a file and can't be removed.
<img width="851" alt="image" src="https://github.com/aws-deadline/deadline-cloud-for-blender/assets/16008714/ea5f2beb-4ff0-4979-be7b-f4452054d182">
Job submission fails:
<img width="372" alt="image" src="https://github.com/aws-deadline/deadline-cloud-for-blender/assets/16008714/f1b7377f-05db-40df-91b7-7bd967359ff6">
The folder cannot be removed from the file list using the "Remove Selected" button:
<img width="372" alt="image" src="https://github.com/aws-deadline/deadline-cloud-for-blender/assets/16008714/ffaafa9b-f3ab-4544-869d-e29e6baa5682">

### Reproduction Steps

- Create a scene with a fluid simulation
- Bake the scene
- Attempt to submit to Deadline Cloud

### Code Snippet

N/A

epmog:
Sorry for the delay, thanks for the report!

The problematic code appears to be here: https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py#L60-L65

`find_files` actually returns folders and files - https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/blender_utils.py#L69-L70

So we should be handling this better.

As a workaround, you should be able to use the `Export Bundle`, modify the `asset_references.yaml` file to remove the reference, and submit the bundle with `deadline bundle submit <path/to/exported/bundle>

epmog:
This should be resolved once https://github.com/aws-deadline/deadline-cloud-for-blender/pull/129 is released

PR #129, opened 2024-10-08
fix: Classify auto-detected assets correctly

miabatta:
# What was the problem/requirement? (What/Why)
A utility function classified all auto-detected assets as files even when some were directories, causing jobs to not submit. Also see issue #105

Created a new function to sort each returned path into the proper set (filenames or directories).

While testing, there were some linting-related issues- we determined that Blender 3.6 uses Python 3.10, so we don't need to test on Python 3.9. Instead the testing matrix should use 3.10, 3.11, and 3.12.

image

Users won't have to manually remove auto-detected assets and can submit jobs as expected.

Manual testing both on my local machine and in a sample customer workspace; job submission succeeds in both cases.

I wrote a simple regression test to make sure files and directories are sorted correctly.

No

No

----

### DIFF
--- a/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
+++ b/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
@@ -9,11 +9,11 @@ from deadline.client import api
 from deadline.client.job_bundle._yaml import deadline_yaml_dump
 from deadline.client.job_bundle.submission import AssetReferences
 from deadline.client.ui.dialogs.submit_job_to_deadline_dialog import SubmitJobToDeadlineDialog
-from deadline_cloud_blender_submitter import blender_utils as bu
-from deadline_cloud_blender_submitter import sanity_checks as sc
-from deadline_cloud_blender_submitter import scene_settings_widget as ssw
-from deadline_cloud_blender_submitter import template_filling as tf
-from deadline_cloud_blender_submitter._version import version_tuple as adaptor_version_tuple
+from . import blender_utils as bu
+from . import sanity_checks as sc
+from . import scene_settings_widget as ssw
+from . import template_filling as tf
+from ._version import version_tuple as adaptor_version_tuple
 
 from qtpy.QtCore import Qt  # type: ignore
 
@@ -54,15 +54,11 @@ def create_deadline_dialog(parent=None) -> SubmitJobToDeadlineDialog:
     # Load and set sticky settings, if any.
     settings.load_sticky_settings(settings.project_path)
 
-    settings.current_layer_selectable_cameras = settings.camera_selection
-    settings.all_layer_selectable_cameras = settings.camera_selection
+    settings.current_layer_selectable_cameras = [settings.camera_selection]
+    settings.all_layer_selectable_cameras = [settings.camera_selection]
 
     # Set auto-detected attachments.
-    files = bu.find_files(settings.project_path)
-    auto_detected_attachments = AssetReferences(
-        input_filenames=set(str(f) for f in files),
-        # input_directories=set(str(f.parent) for f in files),
-    )
+    auto_detected_attachments = _get_auto_detected_assets(settings.project_path)
 
     # Set regular attachments.
     attachments = AssetReferences(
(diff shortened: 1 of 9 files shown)


pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "deadline-cloud-for-blender"
 dynamic = ["version"]
 readme = "README.md"
 license = "Apache-2.0"
-requires-python = ">=3.9"
+requires-python = ">=3.10"
```
jusiskin: I think this would warrant marking this PR as a breaking change. People who were installing the Blender adaptor on Python 3.9 will no longer be able to.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
```diff
@@ -9,11 +9,11 @@
...
-from deadline_cloud_blender_submitter import blender_utils as bu
-from deadline_cloud_blender_submitter import sanity_checks as sc
-from deadline_cloud_blender_submitter import scene_settings_widget as ssw
-from deadline_cloud_blender_submitter import template_filling as tf
-from deadline_cloud_blender_submitter._version import version_tuple as adaptor_version_tuple
+from . import blender_utils as bu
+from . import sanity_checks as sc
+from . import scene_settings_widget as ssw
+from . import template_filling as tf
+from ._version import version_tuple as adaptor_version_tuple
```
amarsjac, line 16: Do you know why the imports needed to be reconfigured?

test/unit/deadline_submitter_for_blender/test_submitter.py
```diff
@@ -6,6 +6,12 @@
 from pathlib import Path
 
 import pytest
+from unittest.mock import Mock, patch
+
+from .mock_bpy import bpy_module as bpy
```
epmog: do we need this mock_bpy (or that file in general) if we're mocking out the module in the `__init__.py`?

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
miabatta: The underlying reason, no. What I encountered was that the `deadline_cloud_blender_submitter` module couldn't be discovered by unit tests, before this change none of the tested modules imported from it. I can get you the error message in a second if you'd like

miabatta:
```
    from deadline_cloud_blender_submitter import template_filling as tf
E   ModuleNotFoundError: No module named 'deadline_cloud_blender_submitter'
```
is from the test logs.

test/unit/deadline_submitter_for_blender/__init__.py
```diff
@@ -1 +1,20 @@
...
+
+import sys
+from unittest.mock import MagicMock
+
+# Mock out UI code to test functions that use the submitter dialog.
+mock_modules = [
+    "PySide2",
+    "PySide2.QtCore",
+    "PySide2.QtGui",
+    "PySide2.QtWidgets",
```
epmog, line 11: PySide6? Or is this not needed with qtpy?

miabatta: Tests run fine without it so I'm led to believe it's not needed!

test/unit/deadline_submitter_for_blender/test_submitter.py
miabatta: You're right, I'd copied this structure from Houdini but the tests run fine with `bpy` mocked out in `__init__.py`. That'll be in my next revision

jusiskin (changes requested):

pyproject.toml
epmog: 🤦 Not sure why I completely glossed over when talking this through earlier. I believe it was just a type definition that prompted this discussion. Should be simple to change `str | None` to `Optional[str]` to keep it non-breaking and leave the 3.9 -> 3.10 documentation alone

miabatta: I'll keep 3.12 in but also add 3.9 back, if I run into syntax issues it shouldn't be difficult to take care of

epmog (approved):

amarsjac (approved):

jusiskin (approved):

Issue #136, opened 2024-10-24
Bug: Output path is ignored and over written in submitter


lunacava:
### Expected Behaviour

"output path" in submitter is set as the setting render output directory setting in the render tab.

### Current Behaviour

when I submit from Blender, Output path is over written in submitter, ignored what I put in output directory in render tab.
![image (37)](https://github.com/user-attachments/assets/1282dac9-24c0-483a-af27-4d23cf893e1a)

### Reproduction Steps

make a scene,
set render setting, put render output directory in Render tab,
open Deadline Cloud render submitter,
go to Job-specific settings,
output pass is not set as I put in the render tab.
and I need to reset output pass every time submit the job. 

### Code Snippet

*

miabatta:
Hi! Thanks for letting us know about this. I'll be putting up a pull request to fix the bug shortly.

miabatta:
Fixed in the change above. I also noticed you're specifying an output filename in that scene, which the submitter also doesn't support yet- working on adding that next!

PR #144, opened 2024-11-13
fix: Use user-specified output directories when applicable

miabatta:
Fixes (some of) https://github.com/aws-deadline/deadline-cloud-for-blender/issues/136.

Blender provides users two ways to specify output directories (In Preferences > File Paths > Render Output, or in Scene > Output). However, the submitter overwrites these options and sets the output directory to be the whichever directory the scene file is in.

Check if the scene has an output path set, and use the directory if it exists. If not, look for a render output directory in preferences. Only if neither of these exist do we set the output directory in the submitter dialog to be the scene file directory. The Blender APIs in this change exist in both Blender 3.6 and 4.2.

File paths are converted to absolute paths before submission- sometimes Blender uses relative paths that don't map nicely to sessions :) (I can elaborate if necessary)

Users will not have to manually override the output directory value after already setting it in Blender.

Note that custom-specified output file names are still not supported and will use a default value. This will be addressed soon in a future fix.

Manually tested with different values in the scene output and preferences. Ran existing unit tests to confirm nothing broke.

No new unit tests have been added since they would just mock the Blender API and test if-statements.

No

No

----

### DIFF
--- a/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/sanity_checks.py
+++ b/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/sanity_checks.py
@@ -4,6 +4,7 @@
 Sanity checks done on submit or export bundle
 """
 from typing import Union
+from pathlib import Path
 
 import bpy
 from qtpy import QtWidgets
@@ -64,6 +65,15 @@ def run_sanity_checks(settings):
             "Please select a different ViewLayer.".format(settings.view_layer_selection)
         )
 
+    # Ensure the specified output directory exists.
+    # If not, create it
+    output_dir_as_path = Path(settings.output_path)
+    if not (output_dir_as_path.exists()):
+        try:
+            output_dir_as_path.mkdir(parents=True)
+        except Exception as e:
+            raise RuntimeError(f"Could not create {output_dir_as_path}.\n\nError message:\n{e}")
+
 
 def _prompt_unsaved_changes():
     """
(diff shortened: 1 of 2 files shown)


src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/sanity_checks.py
```diff
@@ -64,6 +65,16 @@ def run_sanity_checks(settings):
...
         )
 
+    # Ensure the specified output directory exists
+    output_dir_as_path = Path(settings.output_path)
+    if not (output_dir_as_path.exists() and output_dir_as_path.is_dir()):
+        raise RuntimeError(
+            "The output directory ({}) does not exist.\n"
+            "Please choose an output directory in the submitter dialog, the scene settings, or your preferences.".format(
+                settings.output_path
+            )
```
epmog: This is after we've grabbed the output location from the scene/preferences, and after a user has specified a folder path? 

This seems like something we can (should?) solve for the user by creating the folder, assuming we have a path they've specified. Thoughts?

miabatta: Sure thing, that makes sense. I'd add a confirmation dialog to this effect after the user hits submit :)

epmog (approved):
Ship it!

lucaseck (approved):

PR #145, opened 2024-11-18
fix: Apply output file prefixes to submitter GUI

### LINKED ISSUE #136: Bug: Output path is ignored and over written in submitter

miabatta:
Fixes https://github.com/aws-deadline/deadline-cloud-for-blender/issues/136.

Users can specify output file prefixes in the Blender UI, but they are not shown in the submitter GUI. The `OutputFileName` job parameter is hardcoded to a default value.

Addeda widget to the integrated submitter for Output File Prefix. On opening the submitter, these look for a file prefix set in the Blender UI. It will display the job template's default prefix, `output_####`, if the value is unset.

Users can save render output with different file names using the Blender submitter.

Edited existing unit tests to include the new UI settings values & ran them successfully.

I manually tested the adaptor changes to make sure render output formats matched those specified in the submitter.

Changes involving the Blender UI were manually tested using different combinations of input. I'd recommend automated UI regression tests in the future to make sure no job template parameters are left out of the integrated submitter UI.

No

**Edit 03/12/24**: No

----


src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
```diff
@@ -52,13 +52,38 @@ def create_deadline_dialog(parent=None) -> SubmitJobToDeadlineDialog:
...
     else:
-        settings.output_path = os.path.dirname(bpy.context.blend_data.filepath)
+        settings.output_directory = os.path.dirname(bpy.context.blend_data.filepath)
+
+    if bpy.path.basename(bpy.context.scene.render.filepath):
+        settings.output_file_prefix = bpy.path.basename(bpy.context.scene.render.filepath)
+
+    # When setting the output format, we can only access the extension from the Blender API.
+    # Here, we map them to formats available in the submitter GUI combobox.
+    file_formats = {
```
lucaseck: Could a mapping be created like was done previously for KeyShot where there was a [similar issue](https://github.com/aws-deadline/deadline-cloud-for-keyshot/pull/3/files)? That way the definitions of the mappings only live in one place.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
```diff
@@ -52,13 +52,38 @@ def create_deadline_dialog(parent=None) -> SubmitJobToDeadlineDialog:
...
+        ".jp2": "JP2",
+        ".tga": "TGA",  # Can represent both TGA and RAWTGA - we leave it up to the user to specify
+        ".cin": "CINEON",
+        ".dpx": "DPX",
+        ".exr": "OPEN_EXR",  # Again, could be both OPEN_EXR and OPEN_EXR_MULTILAYER
+        ".hdr": "HDR",
+        ".tif": "TIFF",
+        ".webp": "WEBP",
+    }
+    try:
```
lucaseck: nit: Could simplify this a little with `settings.output_format = file_formats.get(bpy.context.scene.render.file_extension, "PNG")`

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/template_filling.py
```diff
@@ -49,7 +49,9 @@ class BlenderSubmitterUISettings:
 
     # Paths and files.
     project_path: str = field(default="")
-    output_path: str = field(default="")
```
lucaseck, line 52: Does this need to be a breaking change by doing this rename?

lucaseck (commented):
Thanks for taking this on Mia! I have one main concern where I don't think the adaptor is ever doing anything with the output format right now [here](https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py#L175) 

It looks like the adaptor will still be [constructing the output path](https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py#L115) with the directory/layer_camera_prefix, but never adding the format? 

The layer and camera being added into the output path automatically is probably something we want to make configurable/optional as well, but doesn't need to be included with this change in my opinion.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
miabatta: Good call, will fix!

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
miabatta: Definitely, that'll be a quick fix

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/template_filling.py
miabatta: For visibility: we talked about this offline and the reasoning is that this rename is a breaking change for people building off this repository.

My preference for `directory` over `path` is purely semantic so we can keep the name as-is, as long as it's clear that it's the path to the directory :)

miabatta:
> I have one main concern where I don't think the adaptor is ever doing anything with the output format right now [here](https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py#L175)
> 
> It looks like the adaptor will still be [constructing the output path](https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py#L115) with the directory/layer_camera_prefix, but never adding the format?

Thank you so much for bringing this up @lucaseck - I should have noticed, but hadn't tried overriding the scene value in the submitter and verifying the final format.

It shouldn't be a huge adaptor change to include this. I'll make these changes and let you know when they're ready for review!

erico-aws (dismissed):

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
```diff
@@ -115,6 +115,11 @@ def start_render(self, data: dict) -> None:
             f"{self.output_dir}/{self.view_layer_name}_{camera}_{self.output_file_name}"
         )
         _logger.debug(f"Set output file path to {bpy.context.scene.render.filepath}")
+
+        # Add the output format
+        bpy.context.scene.render.image_settings.file_format = self.format
```
lucaseck: nit: The adaptor schema isn't an enum so I'd be curious what happens if someone isn't using the submitter but is using the adaptor and their own template that isn't a "valid" file format.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
```diff
@@ -60,6 +60,17 @@ def create_deadline_dialog(parent=None) -> SubmitJobToDeadlineDialog:
     else:
         settings.output_path = os.path.dirname(bpy.context.blend_data.filepath)
 
+    if bpy.path.basename(bpy.context.scene.render.filepath):
+        settings.output_file_prefix = bpy.path.basename(bpy.context.scene.render.filepath)
+
+    # Not all output formats in the Blender UI are supported by the adaptor,
```
lucaseck: Why aren't they all supported? What about them makes them not supported.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/scene_settings_widget.py
```diff
@@ -32,6 +32,22 @@
 COMBO_DEFAULT_ALL_RENDERABLE_CAMERAS = "All Renderable Cameras"
 COMBO_DEFAULT_ALL_RENDERABLE_LAYERS = "All Renderable Layers"
 
+_BLENDER_OUTPUT_FORMATS: dict[str, str] = {
```
lucaseck: Does this need to be private?

miabatta: I don't think it *needs* to. I was following Keyshot which has this dictionary private but I can make it public if that might help future work.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
miabatta: I might have been mistaken here; the ones that I thought weren't supported are all video encodings. I based this assumption off of the original job template's allowed values, which we now know didn't do anything...

I'll run some tests with video encodings & see if there's any reason not to include them!

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
miabatta: I'll need to make some changes anyway so I can test this out + add validation on the adaptor's end if necessary.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
lucaseck: Ah that makes sense. Not sure how the video encodings will work right now. My guess is we would be left with a bunch of single frame videos. If they are all video only ones right now I'd say leave them out.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/scene_settings_widget.py
lucaseck: I guess what I was getting at is this is being used here: https://github.com/aws-deadline/deadline-cloud-for-blender/pull/145/files/106a3ceb1de0bd22dbd1dea0eaa489db68ee6f94#diff-afa6ff66d18ce2a86bdc1eeef3e4b44ad32dfe76e5e50a3355123cfa82346a2fR69

Would it be any cleaner to move that check within the widget file here somewhere? The format can be swapped to the default of PNG likely on first load if it's not in the dict all within this file...I think.

miabatta: Oh, that should be fine I believe. I'll make a separate function in `scene_settings_widget.py` for the validation

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
miabatta: Just confirmed through testing that trying to render frames 1-3 as a video results in 3 one-frame videos. I will keep this as-is!

lucaseck (approved):

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/scene_settings_widget.py
lucaseck: Chatted offline with Mia and the output format handling will be a separate PR

erico-aws (approved):

PR #156, opened 2024-12-03
feat(adaptor): Add BLENDER_EXECUTABLE environment variable

Ahuge:

The assumption that the blender executable exists on the PATH doesn't always make sense, often it is easier to specify the version of Blender to use by explicit path and using an environment variable is an easy way to do this for customers.

We have defined a new environment variable BLENDER_EXECUTABLE which follows the naming from 
[`deadline-cloud-for-cinema-4d`](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/mainline/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py#L298) of `<EXE_NAME>_EXECUTABLE`
Additional functionality, non-breaking, adds the environment variable override but falls back to existing functionality
Yes added in the README.md

No
----

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,10 @@ The Blender Adaptor implements the [OpenJD][openjd-adaptor-runtime] interface th
 * a standardized render application interface,
 * sticky rendering, where the application stays open between tasks,
 
-Jobs created by the submitter use this adaptor by default.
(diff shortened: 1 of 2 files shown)


lucaseck (approved):

epmog (approved):

Ahuge:
Rebased off of mainline

PR #158, opened 2024-12-10
fix: Add OCIO configuration support in SMF

### LINKED ISSUE #143: Bug: color management information is not captured in Job Attachments
### Expected Behaviour

It would be great if *somehow* the submitter could determine what color space information ( e.g. configs) are being used in the scene, and then store them as attachments, so that when rendering in SMF color management environments can be duplicated.

Unfortunately there are a number of ways to set this up!
- centralised data referenced by OCIO environment variables
- data dumped into the install location of Blender.

A workaround would be to add something to the queue to pass the OCIO folder , and then assume the config lies within it, and set OCIO - but not sure if this will scale for all use cases ( more complex hierarchical setups, non-centralised locations of data,etc.)

```
{
	"specificationVersion": "environment-2023-09",
	"parameterDefinitions": [
		{
			"name": "ocio_location",
			"type": "PATH",
			"default": "Z:/OCIO/OpenColorIO-Config-ACES-1.2/aces_1.2",
			"dataFlow": "IN",
			"objectType": "DIRECTORY"
		}
	],
	"environment": {
		"name": "ocio_location",
		"script": {
			"actions": {
				"onEnter": {
					"command": "{{Env.File.Enter}}"
				}
			},
			"embeddedFiles": [
				{
					"name": "Enter",
					"type": "TEXT",
					"runnable": true,
					"data": "#!/bin/bash\necho 'openjd_env: OCIO={{Param.ocio_location}}/config.ocio'"
				}
			]
		}
	}
}
```

### Current Behaviour

No OCIO is captured and scenes are loaded with stock Blender color management.

This results is potential render changes.

### Reproduction Steps

1. Setup OCIO

https://opencolorio.readthedocs.io/en/latest/quick_start/for_artists.html

2. Start Blender and start an example scene.

I love Daniel Bystedt's Splash Fox scene:
https://cloud.blender.org/p/gallery/5f4cd4441b96699a87f70fcb

3. Adjust the scene

Within the shading section , adjust the texture nodes to use 'Utility - sRGB - Texture' 

4.Render and compare.

Render locally, render in SMF, compare the differences.
Write out PNG and EXR files.

### Code Snippet

n/a

miabatta:
Fixes #143.

Users can specify colour management information with the `OCIO` environment variable. However, the contents of this variable are not automatically set on SMF, so a user may unintentionally render something that doesn't use their proper colour management information.

Add functions to upload the OCIO config file, along with any directories referenced in this file, as job attachments so they can be accessed by SMF workers.

The `OCIO` environment variable is set on the worker by adding it in a job environment to the default template. Any additional directories that are added are shown in the submitter dialog as auto-detected assets.

Users can use OCIO in their Blender renders without special set-up.

- Added a unit test that an extra job environment & parameter value are added when OCIO is set in the submitter settings.
- Tested rendering with:
  - No OCIO environment variable
  - `OCIO` set to the default Blender colour management data, which references two other directories
  - `OCIO` set to an [example configuration from the Academy Software Foundation](https://github.com/AcademySoftwareFoundation/OpenColorIO-Config-ACES/releases) (`cg-config-v2.1.0_aces-v1.3_ocio-v2.2.ocio`)

Note that, when testing, Blender 3.6 only supports OCIO versions <= 2.2.

The manual test results are attached! `no_setting` and `default` are expected to look the same because they both use the default Blender config. We can confirm that the right file was uploaded by looking at the task logs.

No OCIO set:
RenderLayer_Camera_test_ocio_no_config0001

Default OCIO uploaded:
RenderLayer_Camera_test_ocio_default_config0001

With this line from the task logs:
`2024/12/10 16:33:33-06:00 ADAPTOR_OUTPUT: STDOUT: Color management: Using /sessions/[session ID]/[asset root]/Program Files/Blender Foundation/Blender 3.6/3.6/datafiles/colormanagement/config.ocio as a configuration file`

ACES configuration uploaded:
RenderLayer_Camera_ocio_file_example0001

No
No

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,7 +91,7 @@ explicit_package_bases = true
 mypy_path = "src"
 
 [[tool.mypy.overrides]]
-module = ["bpy.*", "qtpy.*"]
+module = ["bpy.*", "qtpy.*", "PyOpenColorIO.*"]
 
 # --- RUFF / BLACK ---
 
--- a/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/ocio_utils.py
+++ b/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/ocio_utils.py
@@ -1,0 +1,32 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+import os
+from pathlib import Path
+
+import PyOpenColorIO as ocio
+
+
+def get_ocio_path() -> str:
+    return os.environ["OCIO"] if "OCIO" in os.environ else ""
+
+
+def get_ocio_config(filename: str) -> ocio.Config:
+    return ocio.Config.CreateFromFile(filename)
+
+
+def get_ocio_referenced_dirs(config: ocio.Config) -> list[str]:
+    """
+    Given an OCIO Config object, parse its `search_path` attribute and
+    return a list of referenced directories.
+    """
+
+    path_list = []
+    if config.getSearchPath():
+        for path in config.getSearchPaths():
+            # Paths may be absolute or relative to the config file.
+            if not Path(path).is_absolute():
+                path = Path.joinpath(Path(os.path.dirname(get_ocio_path())), path)
+
+            path_list.append(path)
+
+    return path_list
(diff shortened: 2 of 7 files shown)


rena-cheng:
Nice!

leongdl:
General question - can we please check in a blender template example with OCIO? This would be a great addition to the automated integration testing we add.

leongdl (approved):

rena-cheng (approved):

PR #159, opened 2024-12-13
feat!: Enable GPU rendering from the submitter

### LINKED ISSUE #148: Bug: Submitter does not copy preference resulting in Cycles GPU renders running as CPU in a GPU (SMF) Fleet
### Expected Behaviour

The preferences should be added to job attachments, so that the Cycles Render Devices in the System Preferences are preserved, and stipulate GPU rendering.

For a workaround, you can add this to the Queue Environment :

```
{
	"specificationVersion": "environment-2023-09",
	"parameterDefinitions": [
		{
			"name": "config_location",
			"type": "PATH",
			"default": "C:/Users/Administrator/AppData/Roaming/Blender Foundation/Blender/4.2/config",
			"dataFlow": "IN",
			"objectType": "DIRECTORY"
		}
	],
	"environment": {
		"name": "config_location",
		"script": {
			"actions": {
				"onEnter": {
					"command": "{{Env.File.Enter}}"
				}
			},
			"embeddedFiles": [
				{
					"name": "Enter",
					"type": "TEXT",
					"runnable": true,
					"data": "#!/bin/bash\necho 'openjd_env: BLENDER_USER_CONFIG={{Param.config_location}}/'"
				}
			]
		}
	}
}
```

Note that for some reason passing in a convenient value such as %APPDATA% on Windows to derive the preferences does not evaluate correctly, and the location of %APPDATA% gets appended to the Program Files folder, so you have to explicitly declare a good default.

### Current Behaviour

No preferences are passed, resulting in a (default) CPU render.

### Reproduction Steps

Setup a scene to render GPU:
- edit the Preferences, and select CUDA with a GPU
- in the render properties of the scene, selec the cycles render engine, and specify GPU Compute.

Render on a GPU farm vai Deadline Cloud - I can't see any direct warnings about this, but you can derive that its CPU based on the excessive render times .

### Code Snippet

n/a

miabatta:
Fixes #148. 

### Edits 02/01/25:
- Changed the new parameter's name to `gpu_device`.
- Exposed GPU rendering & device selection in the submitter UI (auto-populated based on the user & scene settings).

image

- Adaptor now iterates through Blender's devices if the supplied GPU device isn't available.

image

- Other misc. code refactoring as suggested!

Nothing in the submitter/adaptor enables GPU rendering based on the user's preferences or scene settings, meaning scenes with GPU rendering enabled will still render using the CPU on SMF.

- Add the parameter `GPUDevice` to the default job template. This has a default value of `NONE` (i.e., use the CPU) and accepts any value.
- Edited the adaptor schema to have a `gpu_device` attribute.
- Added a new action type, `set_gpu_device`, to the adaptor.
- Allow users to choose from Blender's natively-supported GPU device types, or manually enter their own. If they have enabled GPU rendering but their chosen device isn't available on the worker, the worker will iterate through Blender's supported device types until it finds one with compatible devices. Fall back to CPU rendering IFF no supported GPU devices are available.

GPU rendering is available through the Blender submitter without extra setup!

- Added/updated submitter unit tests to account for the new parameter.
- Manually tested GPU renders with all four device types on both 3.6 and 4.2. Compare the times to render 3 frames:

image
image

Relevant docstrings were updated.

**Yes**, this adds a new attribute to the adaptor schema.

----

### DIFF
--- a/src/deadline/blender_adaptor/BlenderAdaptor/schemas/init_data.schema.json
+++ b/src/deadline/blender_adaptor/BlenderAdaptor/schemas/init_data.schema.json
@@ -12,6 +12,9 @@
                 "cycles"
             ]
         },
+        "gpu_device": {
+            "type": "string"
+        },
         "render_scene": {
             "type": "string"
         },
--- a/src/deadline/blender_adaptor/BlenderClient/render_handlers/cycles_blender_handler.py
+++ b/src/deadline/blender_adaptor/BlenderClient/render_handlers/cycles_blender_handler.py
@@ -1,7 +1,56 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
+import bpy
+import logging
+
 from .default_blender_handler import DefaultBlenderHandler
+
+_logger = logging.getLogger(__name__)
 
 
 class CyclesHandler(DefaultBlenderHandler):
     RENDER_ENGINE_NAME = "CYCLES"
+
+    def set_gpu_device(self, data: dict) -> None:
+        """
+        Enables GPU rendering on the worker if specified in the user's configuration.
+
+        Args:
+            data (dict): The data given from the Adaptor. Keys expected: ['gpu_device']
+
+        Raises:
+            RuntimeError: If no devices of the specified type are available
+        """
+
+        gpu_device = data.get("gpu_device", "NONE")
+        if gpu_device == "NONE":
+            bpy.context.scene.cycles.device = "CPU"
+            return
+
+        # Supported GPU devices for Blender 3.6 and 4.2:
+        # https://docs.blender.org/manual/en/3.6/render/cycles/gpu_rendering.html
+        # https://docs.blender.org/manual/en/4.2/render/cycles/gpu_rendering.html
+        GPU_DEVICE_TYPES_ITER = iter(["OPTIX", "CUDA", "HIP", "ONEAPI", "METAL"])
+
+        cycles_preferences = bpy.context.preferences.addons["cycles"].preferences
+        cycles_preferences.refresh_devices()
+
+        # Attempt to set the user's chosen render device.
+        # If it isn't available, loop through device types to find another GPU to use.
+        devices = self._list_compatible_render_devices(cycles_preferences, gpu_device)
+        while not devices and gpu_device != "NONE":
+            _logger.warning(f'Device type "{gpu_device}" is not available on this worker.')
+            gpu_device = next(GPU_DEVICE_TYPES_ITER, "NONE")
+            _logger.warning(f'Attempting to use "{gpu_device}" instead.')
+            devices = self._list_compatible_render_devices(cycles_preferences, gpu_device)
+
+        bpy.context.scene.cycles.device = "GPU" if gpu_device != "NONE" else "CPU"
+        cycles_preferences.compute_device_type = gpu_device
+        _logger.debug(f"Set render device to {gpu_device}.")
+
+    def _list_compatible_render_devices(self, preferences, gpu_device: str) -> list[str]:
+        try:
+            return preferences.get_devices_for_type(gpu_device)
+        except ValueError as ve:
+            _logger.warning(str(ve))
+            return []
--- a/src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
+++ b/src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
@@ -21,6 +21,7 @@ class DefaultBlenderHandler:
         # These cross-reference the Action instances created and queued in the Blender adaptor.
         self.action_dict = {
             "scene_file": self.set_scene_file,
+            "gpu_device": self.set_gpu_device,
             "render_scene": self.set_render_scene,
             "view_layer": self.set_view_layer,
             "camera": self.set_camera,
@@ -218,3 +219,10 @@ class DefaultBlenderHandler:
         if not os.path.isfile(file_path):
             raise FileNotFoundError(f"The scene file '{file_path}' does not exist")
         bpy.ops.wm.open_mainfile(filepath=file_path)
+
+    def set_gpu_device(self, data: dict) -> None:
+        """
+        Only implemented in the Cycles handler.
+        """
+
+        pass
--- a/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/default_blender_template.yaml
+++ b/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/default_blender_template.yaml
@@ -14,7 +14,7 @@ parameterDefinitions:
       patterns: ["*.blend"]
     - label: All Files
       patterns: ["*"]
-  description: Choose the Blender scene file you want to render.
+  description: The Blender scene file you want to render.
 - name: RenderEngine
   type: STRING
   default: cycles
@@ -32,7 +32,7 @@ parameterDefinitions:
   userInterface:
     control: LINE_EDIT
     label: view_layer
-  description: Choose the layer to render.
+  description: The layer to render.
   default: ViewLayer
 - name: Frames
   type: STRING
@@ -49,22 +49,29 @@ parameterDefinitions:
   userInterface:
     control: CHOOSE_DIRECTORY
     label: Output Directory
-  description: Choose the render output directory.
+  description: The render output directory.
 - name: OutputFileName
   type: STRING
   userInterface:
     control: LINE_EDIT
     label: Output File Name
   default: "output_####"
-  description: Enter the output filename (without extension).
+  description: The output filename (without extension).
 - name: OutputFormat
   type: STRING
   userInterface:
     control: DROPDOWN_LIST
     label: Output File Format
-  description: Choose the file format to render as.
+  description: The file format to render as.
   default: PNG
   allowedValues: [TARGA, TARGA_RAW, JPEG, IRIS, PNG, HDR, TIFF, OPEN_EXR, OPEN_EXR_MULTILAYER, CINEON, DPX, JPEG2000, WEBP]
+- name: GPUDevice
+  type: STRING
+  userInterface:
+    control: LINE_EDIT
+    label: GPU Device
+  description: The GPU device type to render with when using the cycles engine.
+  default: NONE
 
 - name: StrictErrorChecking
   type: STRING
@@ -95,6 +102,7 @@ steps:
           data: |
             scene_file: {{Param.BlenderFile}}
             render_engine: {{Param.RenderEngine}}
+            gpu_device: {{Param.GPUDevice}}
             render_scene: {{Param.RenderScene}}
             view_layer: {{Param.ViewLayer}}
             output_dir: {{Param.OutputDir}}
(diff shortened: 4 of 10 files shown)


src/deadline/blender_adaptor/BlenderAdaptor/schemas/init_data.schema.json
```diff
@@ -12,6 +12,15 @@
                 "cycles"
             ]
         },
+        "render_device": {
+            "enum": [
+                "NONE",
```
edwards-aws: Since `NONE` in this instance is always going to mean `CPU` what do you think about this being `CPU` instead?

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
```diff
@@ -218,3 +219,35 @@ def set_scene_file(self, data: dict):
...
+        Enables GPU rendering on the worker if specified in the user's configuration.
+
+        Args:
+            data (dict): The data given from the Adaptor. Keys expected: ['render_device']
+
+        Raises:
+            RuntimeError: If no devices of the specified type are available
+        """
+
+        if not data["render_device"] == "NONE":
```
edwards-aws: What do you think about doing an early return here? I generally find less indents easier to read and conditionals without negatives easier to reason about. IMO it also makes it easier to see that when the device is `NONE` nothing is supposed to happen. This is more of a stylistic thing, so it's up to you.

```python
if data["render_device"] == "NONE":
    return

cycles_preferences = bpy.context.preferences.addons["cycles"].preferences
cycles_preferences.refresh_devices()
devices = cycles_preferences.devices

if not devices:
    raise RuntimeError(
        f"Device type {data['render_device']} is not available on this worker."
    )

for device in devices:
    if device.type == "CPU":
        device.use = False
    else:
        device.use = True
        print(f"Enabled GPU device: {device.name}")

cycles_preferences.compute_device_type = data["render_device"]
print(f"Set render device to {data['render_device']}")
bpy.context.scene.cycles.device = "GPU"
```

test/unit/deadline_submitter_for_blender/test_submitter.py
```diff
@@ -368,49 +381,72 @@ def test_get_param_values(submitter_settings, common_layer_settings):
     }
     expected = [{"name": k, "value": v} for k, v in expected_settings.items()]
 
-    filled = template_filling.get_parameter_values(
-        submitter_settings, common_layer_settings, queue_params=[]
-    )
-    assert filled == expected
+    # Patching out the scene cycles device setting will not add the CyclesRenderDevice parameter
+    with patch("bpy.context.scene.cycles.device", "CPU"):
```
edwards-aws, line 385: I know that this was already the state of these tests. But I worry over time this will balloon in size. What do you think about either breaking this out into distinct tests or turning this into a parameterized test? This will make it easier to reason about test failures and we'll set ourselves up well for the future.

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
edwards-aws: If someone is using an older submitter this will result in a `KeyError` and crash the adaptor. I'm not sure if that's the behaviour we want here. If the `render_device` key is missing should we print a warning and keep carrying on?

src/deadline/blender_adaptor/BlenderAdaptor/schemas/init_data.schema.json
miabatta: This is actually an enum that Blender defines in its user preferences:
image

So there's one specific case where the user has GPU set in the scene, but "None" set in preferences, so `NONE` will be added to the job template [in `template_filling` line 405](https://github.com/aws-deadline/deadline-cloud-for-blender/pull/159/files#diff-14207e3cc2d1f53c6ec708b177bf44c8654104a332eb11b6b8c36aca62514b6cR405). 

I agree it's kind of misleading- I could either change the parameter name to `gpu_device`/`render_gpu_device`/etc, or add some more conditional stuff to make sure `NONE` is always discarded as a value. Latter may work better considering your other comment about the key error; I'll change it when I update the PR :)

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
miabatta: Yeah I wouldn't mind! Thank you for the suggestion

test/unit/deadline_submitter_for_blender/test_submitter.py
miabatta: I can definitely do that. As an aside I'm not super happy about the state of these tests in general and might see about refactoring them sometime 🤔

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
miabatta: That seems reasonable. I'm a little concerned about users expecting GPUs to be automatically enabled in earlier submitter versions, then wasting time rendering on CPUs, for which we _could_ fail the job instead...but at the same time, wasting time rendering on CPUs was the default behaviour beforehand, so a warning may be fine. Curious if others have thoughts!

src/deadline/blender_adaptor/BlenderAdaptor/schemas/init_data.schema.json
edwards-aws: Ahhhhh, I see. Ya I think renaming to something like `gpu_render_device` would go a long way to help with the confusion.

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
```diff
@@ -218,3 +219,43 @@ def set_scene_file(self, data: dict):
...
+        # Attempt to set the user's chosen render device.
+        # If it isn't available, loop through device types to find another GPU to use.
+        devices = self._list_compatible_render_devices(cycles_preferences, gpu_device)
+        while not devices and gpu_device != "NONE":
+            _logger.warning(f'Device type "{gpu_device}" is not available on this worker.')
+            gpu_device = next(GPU_DEVICE_TYPES_ITER, "NONE")
+            _logger.warning(f'Attempting to use "{gpu_device}" instead.')
+            devices = self._list_compatible_render_devices(cycles_preferences, gpu_device)
+
+        bpy.context.scene.cycles.device = "GPU"
```
edwards-aws: Should this be `CPU` if the `gpu_device` is `NONE`?

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
```diff
@@ -218,3 +219,43 @@ def set_scene_file(self, data: dict):
...
+        Raises:
+            RuntimeError: If no devices of the specified type are available
+        """
+
+        # Supported GPU devices for Blender 3.6 and 4.2:
+        # https://docs.blender.org/manual/en/3.6/render/cycles/gpu_rendering.html
+        # https://docs.blender.org/manual/en/4.2/render/cycles/gpu_rendering.html
+        GPU_DEVICE_TYPES_ITER = iter(["OPTIX", "CUDA", "HIP", "ONEAPI", "METAL"])
+
+        gpu_device = data.get("gpu_device", "NONE")
```
edwards-aws: Can we do an early exit here if the `gpu_device` is `NONE`? That means we're using a CPU right?

miabatta: Yeah, that was forgetfulness on my part. The check & early return will take care of your other comment re: setting the cycles device to `CPU`.

miabatta:
For visibility: I also talked offline with Cody about putting the adaptor code in the cycles handler rather than the default blender handler. I'm expecting that to be the last major change.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/default_blender_template.yaml
```diff
@@ -65,6 +65,13 @@ parameterDefinitions:
   description: Choose the file format to render as.
   default: PNG
   allowedValues: [TARGA, TARGA_RAW, JPEG, IRIS, PNG, HDR, TIFF, OPEN_EXR, OPEN_EXR_MULTILAYER, CINEON, DPX, JPEG2000, WEBP]
+- name: GPUDevice
+  type: STRING
+  userInterface:
+    control: LINE_EDIT
+    label: GPU Device
+  description: Choose the GPU device type to render with when using the cycles engine.
```
evanspearman-a: Nit: Stylistically I think the word "Choose" should be omitted.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/default_blender_template.yaml
```diff
@@ -65,6 +65,13 @@ parameterDefinitions:
   description: Choose the file format to render as.
   default: PNG
   allowedValues: [TARGA, TARGA_RAW, JPEG, IRIS, PNG, HDR, TIFF, OPEN_EXR, OPEN_EXR_MULTILAYER, CINEON, DPX, JPEG2000, WEBP]
+- name: GPUDevice
+  type: STRING
+  userInterface:
+    control: LINE_EDIT
```
evanspearman-a, line 71: Any reason we're not using a drop down here?

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/default_blender_template.yaml
miabatta: Agreed, I don't mind changing it, especially if it stays a line_edit. As an aside, I could also go in and change the other parameters that use this convention another time.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/default_blender_template.yaml
miabatta: Mostly to be inclusive of other potential setups; I wanted to avoid adding the `ALLOWED_VALUES` attribute if possible, which is required for dropdown parameters. It's a "just-in-case" measure if users have addons/workflows/etc. that would use other GPU device types.

evanspearman-a (dismissed):

src/deadline/blender_adaptor/BlenderClient/render_handlers/cycles_blender_handler.py
```diff
@@ -1,7 +1,56 @@
...
+        # Attempt to set the user's chosen render device.
+        # If it isn't available, loop through device types to find another GPU to use.
+        devices = self._list_compatible_render_devices(cycles_preferences, gpu_device)
+        while not devices and gpu_device != "NONE":
+            _logger.warning(f'Device type "{gpu_device}" is not available on this worker.')
+            gpu_device = next(GPU_DEVICE_TYPES_ITER, "NONE")
+            _logger.warning(f'Attempting to use "{gpu_device}" instead.')
+            devices = self._list_compatible_render_devices(cycles_preferences, gpu_device)
+
+        bpy.context.scene.cycles.device = "GPU"
```
edwards-aws: Do we have an edge case here where the iterator gets to the end and `gpu_device` is `NONE`? I think this should be `CPU` in that case?

miabatta: Good catch, thanks! Should be fast to fix up.

edwards-aws (approved):
Thanks Mia! This looks good to me!

evanspearman-a (approved):

PR #161, opened 2024-12-20
test: add integration tests

edwards-aws:
We need integration tests that test how our submitter and adaptor interact with Blender

Add tests 😄.

This is a first stab at adding integration tests to Blender. You can run all of the tests with `hatch run integ:test`, submitter tests only with `hatch run integ:test_submitter`, and adaptor tests only with `hatch run integ:test_adaptor`.

The submitter tests work by running a script in Blender that generates a job bundle using the the Deadline Cloud addon. We then check the validity of the template and compare the bundle to what we expect.

The adaptor tests work by feeding params and a template to `openjd run` so that we don't have to take care of parsing and whatnot in the tests.

More confidence in our changes.

Ran the tests, I also made some changes to skip the save confirmation when creating a job bundle. I ensured this still showed up when using the UI. 

Yep, added guidelines on running the tests to the `DEVELOPMENT.md`.

Nope.

----


test/integ/helpers/image_comparision.py
```diff
@@ -0,0 +1,24 @@
...
+import numpy as np
+
+
+def assert_all_images_close(expected_image_directory: Path, actual_image_directory):
+    for image in (expected_image_directory).iterdir():
+        if not image.is_file():
+            continue
+
+        # Open the two image files with Pillow https://pillow.readthedocs.io/en/stable/index.html
+        # and put them in numpy arrays. Pillow doesn't have a good built-in way to do image comparision
```
miabatta: spelling nit: comparision -> comparison (also will have to change the file name)

test/integ/test_blender.py
```diff
@@ -0,0 +1,127 @@
...
+    test_file_location = script_location / "minimal_test"
+    scene_location = test_file_location / "scene" / "test.blend"
+    output_path = tmp_path / "output"
+
+    job_params = {
+        "BlenderFile": str(scene_location),
+        "OutputFileName": "image_####.png",
+        "OutputDir": str(output_path),
+        "RenderScene": "Scene",
+        "RenderEngine": "cycles",
```
miabatta, line 117: Is there value in parameterizing the different render engines here? Or if not parameterizing, having different test cases for them? It comes to mind since the output can look quite different from engine to engine.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
```diff
@@ -128,6 +128,33 @@ def _create_bundle(
...
+    job_bundle_dir: str,
+    settings: tf.BlenderSubmitterUISettings,
+    queue_parameters: list[dict[str, Any]],
+    asset_references: AssetReferences,
+    requirements: Optional[dict[str, Any]] = None,
+    purpose: Optional[str] = None,
+    prompt_for_saving=True,
+) -> None:
+    """Create and write Deadline job bundle files to the given directory. Also save sticky settings.
+    The internal variation exists so that the functionality can be accessed without a UI.
```
miabatta, line 164: Great! I'd like to use this to improve the unit tests as well sometime soon.

test/integ/test_blender.py
```diff
@@ -0,0 +1,127 @@
...
+
+from pathlib import Path
+from typing import Any
+
+from .helpers.test_runners import run_adaptor_test, run_blender_submitter_test, is_valid_template
+from .helpers.image_comparision import assert_all_images_close
+
+
+@pytest.mark.submitter
+def test_minimal_scene_submitter(
```
miabatta, line 15: Forgive me if I'm rehashing old discussion here. But some of these tests seem like they could be added to the submitter unit tests. The `openjd check` makes sense in here IMO, but the template filling/parameter validation are covered in `test_submitter.py` (we'd have to add `asset_references` tests). Are there other advantages to having these as integ tests?

miabatta (changes requested):
There are a few things I've asked about, but this is awesome! Super excited about the new tests 👯

test/integ/helpers/image_comparision.py
edwards-aws: Oops, will fix.

test/integ/test_blender.py
edwards-aws: Not right now, as far as I know the only other engine we have to worry about here is Eevee, which is a GPU renderer as far as I know. We can't guarantee that everyone running these tests has a GPU on their machine. Being able to exclude certain tests based on the environment is something we will need to revisit in the future. For example if we want to test integrations with V-Ray, we won't be able to assume that everyone running that tests has the V-Ray plugin.

test/integ/test_blender.py
edwards-aws: The advantage of having these as integration tests is that we are testing how our submitters work with Blender.

Unit tests are great, you can write a lot of them and make them really specific to test our logic. However unit tests mock out the Blender API because the environment they run in doesn't have Blender available to it, or we're looking to test specific behaviour that's challenging to recreate in an integration test.

Another advantage to integration tests is that when Blender releases a new version, we can test that our assumptions are still true about the Blender API and nothing changed.

miabatta: Good point! Thanks for the clarification.

test/integ/test_blender.py
miabatta: I see what you mean! I did try submitting an EEVEE render to a CPU fleet and it rendered successfully (this was my hunch since I can render locally with it), so it is still an option if you think it's useful. (My assumption is that EEVEE/workbench automatically use GPUs whereas Cycles has to be explicitly set to)

test/integ/test_blender.py
```diff
@@ -0,0 +1,127 @@
...
+            "outputs": {
+                "directories": [],
+            },
+            "referencedPaths": [],
+        }
+    }
+
+    with open(job_history_dir / "asset_references.yaml") as actual:
+        actual_asset_reference = yaml.safe_load(actual)
+        # We don't care what order the filenames list is in, so turn it into a set for easier comparision.
```
miabatta: just the one! I don't mind approving though, we can go back and fix it later.

miabatta (dismissed):

test/integ/test_blender.py
edwards-aws: Oh wow I misspelled that again! I'll be back in this code again soon, so I'll make the fix there, or if another reviewer requests changes in this PR.

test/integ/helpers/image_comparison.py
```diff
@@ -0,0 +1,24 @@
...
+        # Open the two image files with Pillow https://pillow.readthedocs.io/en/stable/index.html
+        # and put them in numpy arrays. Pillow doesn't have a good built-in way to do image comparison
+        # with tolerance.
+        actual = np.asarray(PIL.Image.open(actual_image_directory / image.name))
+        expected = np.asarray(PIL.Image.open(image))
+
+        # Check that the two images are the same within a tolerance.
+        # It's normal for there to be noise in an output image, so it is unlikely that two
+        # renders will be exactly the same.
+        assert np.allclose(actual, expected, atol=2)
```
lucaseck, line 24: Just curious, how did you come up with `atol=2` for the tolerance?

test/integ/helpers/test_runners.py
```diff
@@ -0,0 +1,69 @@
...
+    ]
+
+    if additional_args:
+        args.extend(["--", *additional_args])
+
+    return run_command(args)
+
+
+def is_valid_template(template_location: Path) -> bool:
+    output = run_command(["openjd", "check", str(template_location), "--output", "json"])
```
lucaseck, line 46: Is openjd present in the hatch env or is this relying on it being somewhere on the system path?

test/integ/test_scripts/minimal_test/_test_blend.py
```diff
@@ -0,0 +1,75 @@
...
+)
+from qtpy import QtWidgets
+
+from deadline_cloud_blender_submitter.scene_settings_widget import (
+    COMBO_DEFAULT_ALL_RENDERABLE_CAMERAS,
+    COMBO_DEFAULT_ALL_RENDERABLE_LAYERS,
+)
+
+
+def main(job_history_dir: str, output_dir: str):
```
lucaseck, line 20: nit: Can you include a description of what this test covers e.g. multiple layers, multiple cameras, just the basics etc?

DEVELOPMENT.md
```diff
@@ -39,6 +39,15 @@ hatch run fmt
 hatch run all:test
 ```
 
+### Run Integration Tests
+
+1. Set the environment variable `BLENDER_EXECUTABLE` to the location of the Blender application.
```
lucaseck, line 44: Can something be added to the PR template to ask if they've run the integ tests and point to the instructions here.

lucaseck (commented):
Awesome to see @edwards-aws! I've left a couple of comments, but was also curious if you knew anything about https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/setup_integ_env.sh? Not sure if this was a previous attempt at integ tests and is now unused?

test/integ/test_blender.py
```diff
@@ -0,0 +1,127 @@
...
+        )
+        actual_asset_reference["assetReferences"]["inputs"]["filenames"] = set(
+            actual_asset_reference["assetReferences"]["inputs"]["filenames"]
+        )
+        assert actual_asset_reference == expected_asset_references
+
+
+@pytest.mark.xfail(
+    run=True,
+    reason="There's a bug where multiple view layers are rendered when it should only be one.",
```
lucaseck, line 104: Is there a GitHub issue for this? I didn't see one at a quick glance.

test/integ/test_scripts/minimal_test/_test_blend.py
edwards-aws: Can do

test/integ/test_blender.py
edwards-aws: I was gonna cut one after this is merged because this test is the repro case.

DEVELOPMENT.md
edwards-aws: Sure, I was going to wait until we had more. But I guess there's no reason not to add it now.

test/integ/helpers/test_runners.py
edwards-aws: Good catch! I meant to add this to the hatch env and it looks like I forgot.

test/integ/helpers/image_comparison.py
edwards-aws: Mostly a guess, it's low and seems to work. A more scientific way would be to run a couple renders from Blender and see how they usually differ.

edwards-aws:
> Awesome to see @edwards-aws! I've left a couple of comments, but was also curious if you knew anything about https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/setup_integ_env.sh? Not sure if this was a previous attempt at integ tests and is now unused?

I was unaware of this, I might remove it. It looks like it relies on a pretty specific setup with Blender in an S3 bucket.

test/integ/helpers/image_comparison.py
lucaseck: Figured as much, was just curious how much 2 actually translates to in terms of difference since in the numpy docs they're using very small values as the defaults.

leongdl (commented):
Great job! 
Will have a second pass tomorrow ;)

lucaseck (approved):
🚢

test/integ/helpers/image_comparison.py
edwards-aws: I'll look at tightening this up when I add more tests.

miabatta (approved):
🚢

leongdl (approved):

PR #182, opened 2025-02-06
fix: "All Renderable Layers" option renders only one layer per task

### LINKED ISSUE #165: Bug: All View Layers Being Rendered When Only One Should Be Rendered
### Expected Behaviour

When `All Renderable Layers` is selected, each layer should get its own step where each step only renders one of the view layers.

### Current Behaviour

When `All Renderable Layers` is selected, the second step will render all layers.

### Reproduction Steps

* Create a Blender scene with multiple view layers, make the view layers different enough so they can be distinguished.
* In the submitter select All Renderable Layers in the View Layers field.
* Submit the job.
* See that the second step renders the first and second view layer as the combined layers is what gets output
  * Also note in the adaptor logs that it says that the first view layer is also being rendered.

Alternatively run this integration test https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/test/integ/test_blender.py#L107

### Code Snippet

N/A

miabatta:
Fixes #165 .

**Edit 13/02/25**: Simplified the change to no longer be breaking or add new adaptor actions. Now, the frame is set _and_ the view layer is reset in the `start_render` action before the API call to render the scene.

When rendering a scene with multiple view layers, the Blender adaptor will render the specified view layer as well as any view layers that came before it. This causes render outputs to look different than expected and wastes render time on undesired layers.

See the original issue (#165 ) for more context.

- Changed the view layer to set once per `start_render` action, immediately before rendering.
- Specified a single layer in the `bpy.ops.render.render` API call.

These were necessary because certain Blender API calls reset which layers are used or not. We have to set/unset `use` on the appropriate layers _and_ pass the layer name as a parameter in `bpy.ops.render.render` to avoid unnecessary rendering.

Renders with different view layers will look correct & execute more quickly! (also integration tests will pass)

For demonstration, here are the durations of the minimal test scene using the new adaptor (above) and the current adaptor (below):
image

Unit & integration tests, as well as some manual testing to cover what we don't have integration tests for yet. (e.g., checking that layers render properly with different options in the scene and in the submitter)

I also tested the new adaptor with the current submitter to confirm it's not a breaking change.

```
test/integ/test_blender.py::test_minimal_scene_submitter
[gw0] [ 50%] PASSED test/integ/test_blender.py::test_minimal_scene_submitter 
test/integ/test_blender.py::test_minimal_scene_adaptor
[gw0] [100%] PASSED test/integ/test_blender.py::test_minimal_scene_adaptor 

========================================= slowest 5 durations ========================================== 
71.11s call     test/integ/test_blender.py::test_minimal_scene_adaptor
5.45s call     test/integ/test_blender.py::test_minimal_scene_submitter
0.01s setup    test/integ/test_blender.py::test_minimal_scene_adaptor
0.00s setup    test/integ/test_blender.py::test_minimal_scene_submitter
0.00s teardown test/integ/test_blender.py::test_minimal_scene_adaptor
===================================== 2 passed in 77.69s (0:01:17) ===================================== 
```

N/A

No

----


src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
```diff
@@ -65,66 +65,82 @@ def _ensure_camera(self, data: dict) -> str:
...
         # See: https://docs.blender.org/api/current/bpy.ops.render.html#bpy.ops.render.render
-        bpy.ops.render.render(animation=True, scene=self.scene_name)
+        bpy.ops.render.render(scene=self.scene_name, layer=self.view_layer_name, write_still=True)
 
         # This print statement (including flush) is required for Deadline to pick up successful task completion
         # See the Regex callbacks defined at `BlenderAdaptor/adaptor.py:_get_regex_callbacks`.
-        print(f"BlenderClient: Finished Rendering Frame {frame}", flush=True)
+        print(f"BlenderClient: Finished Rendering Frame {self.frame}", flush=True)
+
+    def set_frame(self, data: dict) -> None:
```
justinsaws: Can the type for `data` be scoped down any more? For example: `Dict[str, Any]`

miabatta: My instincts say yes since the adaptor schema should only really be able to take strings as keys. However [OpenJD uses plain dicts in its base classes](https://github.com/OpenJobDescription/openjd-adaptor-runtime-for-python/blob/712e570bd3f4e32356f80943731cf7a73b491f02/src/openjd/adaptor_runtime/adaptors/_base_adaptor.py#L63) so I guess I'm wondering about consistency here. If it's worth it to scope down the type, I can add that as a quick fix in this PR.

justinsaws: Ok, let's leave it to be consistent. Likely something for a longer term refactor to be as specific as possible.

justinsaws (dismissed):

src/deadline/blender_adaptor/BlenderAdaptor/adaptor.py
```diff
@@ -36,6 +36,7 @@ class BlenderNotRunningError(Exception):
 # Order of execution is important.
 _BLENDER_INIT_KEYS = [
     "render_scene",
+    "frame",
```
AWS-Samuel: The code comment on this list is not super helpful to understand the purpose of what `_BLENDER_INIT_KEYS` is intended to accomplish, but the idea behind these `init` action keys is that these are properties which are set once upon initialization in `on_start`. `frame` is typically something which is done during the `on_run` step. 

Looking at the code in `_populate_action_queue` indicates that since `frame` is not part of the `init_data`, we aren't initializing blender with the frame during `on_start`, so we can remove this line

src/deadline/blender_adaptor/BlenderAdaptor/schemas/init_data.schema.json
```diff
@@ -18,9 +18,6 @@
         "render_scene": {
             "type": "string"
         },
-        "view_layer": {
-            "type": "string"
-        },
```
AWS-Samuel: I don't think we need to make this breaking change by removing `view_layer` from `init_data`. 

This will cause templates created with old versions of the blender submitter to fail when running on a blender adaptor with this change. Since `view_layer` can be updated during `on_run`, I don't think there is a downside to keeping this in

src/deadline/blender_adaptor/BlenderAdaptor/schemas/run_data.schema.json
```diff
@@ -10,10 +10,14 @@
...
+        },
+        "view_layer": {
+            "type": "string"
         }
     },
     "required": [
         "frame",
-        "camera"
+        "camera",
+        "view_layer"
```
AWS-Samuel: Why is `view_layer` required? 

This means that each action (task) would need to specify the view layer, but I think there is a reasonable scenario where a customer would want to set this once on the step level.  

By not adding `view_layer` to `required`, then we can support both cases where a customer wants to change the view layer on a specific action, while also supporting the case of using the view layer defined on the step level

src/deadline/blender_adaptor/BlenderAdaptor/adaptor.py
```diff
@@ -386,6 +387,12 @@ def on_run(self, run_data: dict) -> None:
         # Run the camera action.
         self._action_queue.enqueue_action(Action("camera", run_data))
 
+        # Run the frame action.
+        self._action_queue.enqueue_action(Action("frame", run_data))
+
+        # Run the view layer action.
+        self._action_queue.enqueue_action(Action("view_layer", run_data))
```
AWS-Samuel: If you adopt my suggestion of making `view_layer` optional for `run-data`, we will want to add logic here to only optionally queue this action

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
AWS-Samuel: We did we split this off into its own function? I do like that it's more clear in the adaptor where this happens, but as a result we are technically adding a bit of overhead by requiring another request on the adaptor client side for each render.

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
```diff
@@ -65,66 +65,82 @@ def _ensure_camera(self, data: dict) -> str:
         _logger.debug(f"Fetched camera to render: {camera}")
         return camera
 
-    def _get_layer_to_render(self, data: dict) -> Optional[str]:
-        """Gets the view_layer to render from the provided data.
+    def _process_output_filename(self, original_filename: str, frame: int) -> str:
```
AWS-Samuel: I'm surprised this isn't something blender handles itself. What was the behavior before?

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
```diff
@@ -65,66 +65,82 @@ def _ensure_camera(self, data: dict) -> str:
...
         # The actual camera to render is set in the set_camera function,
         # and should have already been called as an Action.
         camera = self._ensure_camera(data)
 
-        # Add layer and camera info to the output file path.
+        # Add layer, camera, and frame info to the output file path.
+
+        # This should be true anyway or the adaptor would have thrown an error earlier.
+        # Asserted here to make the linter happy.
+        assert self.output_file_name is not None
```
AWS-Samuel: In the off chance this assertion fails, can we provide an error message?

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
```diff
@@ -65,66 +65,82 @@ def _ensure_camera(self, data: dict) -> str:
...
         bpy.context.scene.render.filepath = (
-            f"{self.output_dir}/{self.view_layer_name}_{camera}_{self.output_file_name}"
+            f"{self.output_dir}/{self.view_layer_name}_{camera}_{processed_output_filename}"
         )
         _logger.debug(f"Set output file path to {bpy.context.scene.render.filepath}")
 
         _logger.debug(f"Rendering camera: {camera}")
-        # The `animation` flag is required to correctly set the output file name.
+        # To render only single layers, we must set `write_still` to True and `animation` to False (default).
+        # Because we aren't using `animation`, we manually inject the correct frame numbers into the output path.
```
AWS-Samuel: The [blender docs](https://docs.blender.org/api/current/bpy.ops.render.html#bpy.ops.render.render) seem to indicate that simply providing the layer is all that's needed to render a single layer. If so this would be great because then we could re-add `animation=True` and have blender resolve the frame numbers for us in the file path.

Was that not the case?

miabatta: The docs describe the `layer` parameter with "(used only when animation is disabled)" - you can technically include both parameters (`animation=True, layer=self.view_layer_name`) and the output will look correct, but it will still render unnecessary layers (verified thru the adaptor logs). Setting `animation=False, write_still=True` is, from my experimenting, the only way to render specific layers.

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
miabatta: Can do, thanks!

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
miabatta: Mentioned this in the comment about `bpy.ops.render.render`. Blender does not seem to have an API call for filename processing :(

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
miabatta: There's a bit of cascading reasoning that led to this. Something I learned about the Blender API while testing is that, while we set `.use` on individual layers to `True`/`False` to enable rendering of single layers, certain other Blender API calls will reset them to `True` afterwards. This is why I moved `set_view_layer` to run as the very last step before rendering, and also why we have to run it for each task.

Because we can't render frames from the animation range, we have to call `setFrame` instead of using `bpy.context.scene.frame_start=frame`, etc. `setFrame` is one of the API calls that will reset the layers, which is why I moved the frame setting out of the `start_render` action to happen explicitly before `set_view_layer`.

src/deadline/blender_adaptor/BlenderAdaptor/schemas/run_data.schema.json
miabatta: I'm not too adamant about it being required. Thinking about it, it's fine to exclude in the use case where a scene only has one layer in it, but we should then clarify in docs somewhere that multi-layer scenes will encounter this bug unless you set the layers in `run_data`.

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
AWS-Samuel: Can you add some of that context as a code comment? That's very useful information that we will want to make easily discoverable for future contributors.

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
AWS-Samuel: Would https://docs.blender.org/api/current/bpy.types.RenderSettings.html#bpy.types.RenderSettings.frame_path work?

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
AWS-Samuel: Ah, I missed seeing that. Makes sense. Thanks!

src/deadline/blender_adaptor/BlenderAdaptor/schemas/run_data.schema.json
AWS-Samuel: > multi-layer scenes will encounter this bug unless you set the layers in run_data

That's something we can fix on the submitter side, right?

src/deadline/blender_adaptor/BlenderAdaptor/schemas/init_data.schema.json
miabatta: I'd be down to keep it in `init_data`! I've had some difficulties with changing adaptor schema in the past but once I make the necessary edits I'll reevaluate if it has to be breaking or not. Thanks!

src/deadline/blender_adaptor/BlenderAdaptor/adaptor.py
miabatta: Will do, thanks for the heads-up :)

src/deadline/blender_adaptor/BlenderAdaptor/schemas/init_data.schema.json
miabatta: Tested this out & it's backwards-compatible with the current submitter; thanks for the advice!

The output is also correct, though adaptor logs show the extra layer being rendered :)

src/deadline/blender_adaptor/BlenderAdaptor/schemas/run_data.schema.json
miabatta: You're right, thanks! I'll add this and amend the commit.

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
miabatta: Yes! I was looking in the file operations so I didn't catch that one. Thanks for checking!

justinsaws (dismissed):

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
```diff
@@ -65,28 +64,20 @@ def _ensure_camera(self, data: dict) -> str:
...
-        target = data.get("view_layer")
-        if not target:
-            _logger.debug("No view_layer specified in data.")
-            return None
 
-        # In Blender, set the right scene.
-        scene = bpy.data.scenes[self.scene_name]
-        bpy.context.window.scene = scene
+        if self.view_layer_name is None:
+            raise RuntimeError("View layer was not provided in init data")
```
AWS-Samuel: Nit, this error message is not entirely accurate 
```suggestion
            raise RuntimeError("ensure_view_layer must not be called before set_view_layer")
```

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
```diff
@@ -96,31 +87,44 @@ def start_render(self, data: dict) -> None:
...
-        bpy.context.scene.render.filepath = (
-            f"{self.output_dir}/{self.view_layer_name}_{camera}_{self.output_file_name}"
-        )
+        # Add submitter settings to the output filepath.
+        # This includes output directory, view layer, camera, and frame.
+        try:
+            bpy.context.scene.render.filepath = (
+                f"{self.output_dir}/{self.view_layer_name}_{camera}_{self.output_file_name}"
+            )
+        except:
```
AWS-Samuel: nit: a bare except will catch exceptions like `KeyboardInterrupt` or `SystemExit`. In this case it doesn't really matter because we re-raise, but as a best practice we will want at the minimum 
```suggestion
        except Exception:
```

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
AWS-Samuel: Thanks for updating! It's always nice when we can defer this behaviour to the DCC to ensure it's handled as expected

src/deadline/blender_adaptor/BlenderClient/render_handlers/default_blender_handler.py
AWS-Samuel: Actually I think we might not want to raise an exception at all since `view_layer` is not required in `init-data` or `run-data`

AWS-Samuel: In this case we can likely default to whichever layer(s) are set to enabled in the scene, which I think was the previous behavior

miabatta: Sure! Testing it out quickly, passing an empty string for `layer` in the `render` call will by default render every layer in the scene at once, i.e. what the current adaptor does.

As-is this would only be available for people using custom job templates with the adaptor, but if we really find it's necessary we can add the option in the submitter later

AWS-Samuel: Nice, do we still need to get the name of the layer from the scene so that the places we reference `view_layer_name` are properly updated (mainly output paths)?

miabatta: I'm a bit hesitant because if a scene has multiple layers, we'd just be arbitrarily picking one. Old submitter versions will still write layer names, so we should be fine!

AWS-Samuel (approved):

justinsaws (approved):

PR #198, opened 2025-03-03
feat: Add standalone installer

miabatta:
Installers for all of our DCC submitters are coupled together.

- Create a standalone Blender submitter installer that uses the existing installer component.
- Write a Python script to make dev installers for local testing with evaluation versions of InstallBuilder
- Add tests to verify the installation/uninstallation of the component

A Blender-only installer can be created!

Tested the `build_installer.py` script to create installers for different platforms *on* different platforms, then ran them. Also ran the new unit tests

n/a, no submitter/adaptor changes, just infrastructure

Added to `DEVELOPMENT.md` with information about building & testing.

No

----


scripts/build_installer.py
```diff
@@ -0,0 +1,373 @@
...
+            f"/Applications/InstallBuilder Professional {INSTALL_BUILDER_VERSION}/"
+        )
+    elif sys.platform.startswith("win32"):
+        install_builder_path = (
+            f"C:\\Program Files\\InstallBuilder Professional {INSTALL_BUILDER_VERSION}\\"
+        )
+    elif sys.platform.startswith("linux"):
+        install_builder_path = f"/opt/installbuilder-{INSTALL_BUILDER_VERSION}/"
+
+    return install_builder_path
```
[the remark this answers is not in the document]
miabatta: My EOD is right away, so I'll clean this up tomorrow 👍

.github/workflows/build_installer.yml
```diff
@@ -0,0 +1,34 @@
+name: "Build Linux Installer"
```
leongdl: nit on file name, should this be "build_linux_installer.yml", since we will have "build_windows_installer.yml" later? Or even OSX?

scripts/build_installer.py
```diff
@@ -0,0 +1,373 @@
...
+    A structure representing the parameters for integrating a DCC submitter into the InstallBuilder
+    project.
+    """
+
+    name: str
+    """
+    The name of the DCC application this submitter is for.
+
+    This **must** match up exactly with a corresponding value in the `DccSubmitter` enum in the
+    `lib/constructs/Config.ts` file.
```
jericht: Let's remove these file references

.gitignore
```diff
@@ -16,6 +16,7 @@ dependency_bundle/
 dist/
 plugin_env/
 wheels/
+installer/components/
```
epmog, line 19: can we also had a `license.xml` to the ignored files? Or whatever we expect the license for installbuilder to look like?

.github/workflows/build_installer.yml
miabatta: That depends- it might be neater to keep all of the installers as separate steps in one file, in which case we'd still have to change the name to "Build Installer**s**" :) What do you think?

.gitignore
miabatta: Yep, will be in the next fix :)

test/installer/test_installer.py
```diff
@@ -0,0 +1,90 @@
...
+        raise FileNotFoundError(f"Installer not found at '{path}'")
+
+    if not os.access(path, os.X_OK) and not platform.system() == "Darwin":
+        raise PermissionError(f"Installer at '{path}' is not executable")
+
+    yield Path(path).absolute()
+
+
+@pytest.fixture(scope="function")
+def installed(installer_path, tmpdir):
```
lucaseck: Super nit: looks like tmp_path is preferred over tmpdir: https://docs.pytest.org/en/stable/reference/reference.html#tmpdir

test/installer/test_installer.py
```diff
@@ -0,0 +1,90 @@
...
+def test_install(installed: Path):
+    # GIVEN / WHEN
+    uninstaller = "uninstall.app" if platform.system() == "Darwin" else "uninstall"
+    python_dir = installed / "Submitters" / "Blender" / "python"
+
+    # THEN
+    top_level_dir = [f.name for f in installed.iterdir()]
+    assert "Submitters" in top_level_dir
+    assert "installer_version.txt" in top_level_dir
+    assert uninstaller in top_level_dir
```
lucaseck, line 69: Does this pass on Windows for you? I'm seeing failures in my own changes because of the ".exe" extension on `uninstall.exe`

test/installer/test_installer.py
miabatta: Will fix!

test/installer/test_installer.py
miabatta: Good call, I'll add an extra platform-dependent check here.

.github/workflows/build_installers.yml
```diff
@@ -0,0 +1,34 @@
...
+    inputs:
+      branch:
+        required: true
+        type: string
+
+jobs:
+  BuildLinuxInstaller:
+    if: (github.repository == 'aws-deadline/deadline-cloud-for-blender')
+    name: Build Linux Installer
+    runs-on: ubuntu-latest
```
evanspearman-a, line 14: Will an installbuilder installer built on the latest Ubuntu run on older versions that we support? Typically with Linux you have to build on an OS that uses the oldest version of glibc that you intend to support. I'm not sure if installbuilder does anything to mitigate this.

.github/workflows/build_installers.yml
moorec-aws, line 1: Are we not making this a [re-usable workflow](https://github.com/aws-deadline/.github/tree/mainline/.github/workflows)?

miabatta: That's the plan eventually; for now I've kept in here to keep the changes in one place to simplify testing & reviewing.

installer/DeadlineCloudForBlenderSubmitter.xml
```diff
@@ -0,0 +1,288 @@
...
+
+    <!-- Allow Windows installer to run as regular user, since running as Admin is forced by default -->
+    <requireInstallationByRootUser>0</requireInstallationByRootUser>
+    <requestedExecutionLevel>asInvoker</requestedExecutionLevel>
+
+    <readyToInstallActionList>
+        <!-- Cleanup Submitters Directory Prior to Installation -->
+        <showQuestion>
+            <default>yes</default>
+            <text>The Submitters directory exists and will be deleted, backup files if necessary. Do you wish to continue? </text>
```
evanspearman-a: Why do we delete the entire submitters directory? Why not just the blender one?

installer/DeadlineCloudForBlenderSubmitter.xml
```diff
@@ -0,0 +1,288 @@
...
+                    <program>REG</program>
+                    <programArguments>ADD "${key}" /ve /d "${value}" /f</programArguments>
+                </runProgram>
+            </actionList>
+            <parameterList>
+                <stringParameter name="key"/>
+                <stringParameter name="value"/>
+            </parameterList>
+        </actionDefinition>
+        <actionDefinition name="fnAddEnvironmentVariable">
```
evanspearman-a: Could we change the name to `fnAddLinuxEnvironmentVariable`?

installer/DeadlineCloudForBlenderSubmitter.xml
```diff
@@ -0,0 +1,288 @@
...
+                        </setInstallerVariable>
+                    </actionList>
+                    <elseActionList>
+                        <!-- Pass through to InstallBuilder implementation -->
+                        <addEnvironmentVariable name="${name}" value="${value}" scope="${scope}"/>
+                    </elseActionList>
+                </if>
+            </actionList>
+        </actionDefinition>
+        <actionDefinition name="fnAddPathEnvironmentVariable">
```
evanspearman-a, line 68: This function and the next one also look Linux specific, so I think they should have Linux in the name as well.

scripts/build_installer.py
```diff
@@ -0,0 +1,378 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
evanspearman-a, line 1: Could we add typing to the Python code added in this PR. We don't need it on every variable definition, but all function signatures should have it. `def foo() -> None` can be used when there is no return value as well, that way it's explicit that nothing is returned.

scripts/build_installer.py
```diff
@@ -0,0 +1,378 @@
...
+    p = subprocess.Popen(cmd, **kwargs)
+    stdout, stderr = p.communicate()
+    output = stdout.decode("utf-8") + stderr.decode("utf-8")
+    if p.returncode != 0:
+        raise BadRCError(f"Bad rc ({p.returncode}) for cmd '{cmd}': {output}")
+    return output
+
+
+def download_from_s3(bucket_name, key, output_folder):
+    dest_path = os.path.join(output_folder, os.path.basename(key))
```
evanspearman-a: Nit: I'd recommend using `pathlib` over `os.path` for new development.

scripts/build_installer.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+def get_default_installbuilder_location() -> str:
+    """
+    Returns the default location where InstallBuilder Professional will be installed depending on the platform.
+    """
+
+    install_builder_path = ""
+
+    if sys.platform.startswith("darwin"):
```
evanspearman-a: Nit: If you use `platform.system()` you don't need `startswith`.

evanspearman-a: Alternatively, if using that you could just implement this as a dictionary rather than ifs.

scripts/build_installer.py
```diff
@@ -0,0 +1,378 @@
...
+    Used as an error callback in `shutil.rmtree` to address an AccessDenied error that occurs on Windows.
+    If a directory fails to delete, attempt to add write permissions to it and try again.
+    Re-raise the error if it persists.
+    """
+
+    if not os.access(path, os.W_OK):
+        os.chmod(path, stat.S_IWUSR)
+        func(path)
+    else:
+        raise
```
evanspearman-a, line 106: Wouldn't this raise if the permissions are already okay?

scripts/build_installer.py
```diff
@@ -0,0 +1,378 @@
...
+        "--no-cleanup",
+        dest="cleanup",
+        action="store_false",
+        help=("Do not delete the build components folder after completion."),
+    )
+    parser.add_argument(
+        "--platform",
+        required=True,
+        help="The platform to build an installer for",
+        choices=("windows-x64", "linux-x64", "osx"),
```
evanspearman-a: Nit: Should be macos instead of osx. Apple stopped using the term osx when they moved on to macOS 11.

scripts/build_installer.py
```diff
@@ -0,0 +1,378 @@
...
+# See "Supported Platforms" table in https://releases.installbuilder.com/installbuilder/docs/installbuilder-userguide.html
+INSTALLER_FILENAMES = {
+    "windows-x64": "DeadlineCloudForBlenderSubmitter-windows-x64-installer.exe",
+    "linux-x64": "DeadlineCloudForBlenderSubmitter-linux-x64-installer.run",
+    "osx": "DeadlineCloudForBlenderSubmitter-osx-installer.app",
+}
+
+# This is the directory containing the InstallBuilder root .xml component.
+# All file paths in the InstallBuilder component files are relative to this directory
+INSTALL_BUILDER_VERSION = "24.11.1"
```
evanspearman-a: Can we change this into an argument? We don't want to need to update the version in every DCC repository when we update installbuilder.

scripts/build_installer.py
```diff
@@ -0,0 +1,378 @@
...
+        if missing_args:
+            parser.error(
+                "the following arguments are required for non-dev builds: "
+                f"{', '.join(missing_args)}\n"
+            )
+    else:
+        if os.environ.get("CODEBUILD_BUILD_ID") is not None:
+            parser.error("--local-dev-build cannot be used when running in CodeBuild.")
+    with tempfile.TemporaryDirectory() as workdir:
+        run("python -m pip install --upgrade pip")
```
evanspearman-a: I think this should be left up to the user rather than forcing it in the script.

scripts/build_installer.py
```diff
@@ -0,0 +1,378 @@
...
+                missing_args.append(required_arg.argument)
+        if missing_args:
+            parser.error(
+                "the following arguments are required for non-dev builds: "
+                f"{', '.join(missing_args)}\n"
+            )
+    else:
+        if os.environ.get("CODEBUILD_BUILD_ID") is not None:
+            parser.error("--local-dev-build cannot be used when running in CodeBuild.")
+    with tempfile.TemporaryDirectory() as workdir:
```
evanspearman-a, line 215: Generally I would recommend splitting up arguments and application logic into separate functions (or just leave arguments in the `if __name__ == "__main__"` block). That way, if in the future we want to re-use the application logic in another script, this file already functions like a library we can import.

evanspearman-a (changes requested):

scripts/build_installer.py
miabatta: No worries, I can refactor that. The code I added uses `pathlib` whereas existing stuff used `os.path` so I agree it'd be nice to make it consistent :)

scripts/build_installer.py
miabatta: For sure! Thanks for the suggestion

installer/DeadlineCloudForBlenderSubmitter.xml
miabatta: For context, this creates a directory structure with the same shape as the old shared installer, hence the unnecessary `Submitters/Blender` subdirectories.

I'll change this so that the Blender-related files are directly in the `DeadlineCloudForBlenderSubmitter` folder instead of in that directory structure.

scripts/build_installer.py
miabatta: This also bugs me, but InstallBuilder uses `osx` as a constant in its own code & docs. I chose to use `osx` everywhere s.t. it doesn't become complicated to figure out which is used where.

We can change specifically the CLI argument if that's what your main concern is, all it would need is an extra line to convert `macos` to `osx` when we use the argument when calling InstallBuilder.

scripts/build_installer.py
miabatta: Sounds good. IIRC pip would automatically print a warning to upgrade it if it's out-of-date so we should be covered there

scripts/build_installer.py
miabatta: Thanks! There's a lot of platform checking so I'll be sure to consolidate & clean it up.

scripts/build_installer.py
miabatta: Thought process was that if permissions are ok, the callback wouldn't execute for permission-related errors, and if anything else came up we'd just raise the exception like normal.

I could add the `chmod` before each `shutil.rmtree`, it would be a few extra lines but eliminate the need for the callback if it's too risky a way of handling errors.

.github/workflows/build_installers.yml
miabatta: [Ubuntu 6.0.1 and later](https://installbuilder.com/installbuilder-specifications) is called out in the specifications for running installers built by InstallBuilder. They also list their other supported distros & versions on that same page if you're curious!

It's worth noting that you can build installers for any OS no matter which OS you're running `builder` on, so it seems InstallBuilder will account for that caveat by default (as long as it's a supported Linux version).

scripts/build_installer.py
miabatta: Sure thing. The only place the constant is used is for grabbing the default path to InstallBuilder. If we get rid of those default values and make the path a required argument, it would simplify this logic

scripts/build_installer.py
miabatta: Agree, I'll add that in :)

depsBundle.sh
```diff
@@ -5,8 +5,9 @@ python3 depsBundle.py
 
 rm -f dependency_bundle/deadline_cloud_for_blender_submitter-deps-windows.zip
 rm -f dependency_bundle/deadline_cloud_for_blender_submitter-deps-linux.zip
-rm -f dependency_bundle/deadline_cloud_for_blender_submitter-deps-macos.zip
+rm -f dependency_bundle/deadline_cloud_for_blender_submitter-deps-macos*.zip
```
edwards-aws: Is the glob `*` to account for `deadline_cloud_for_blender_submitter-deps-macos-intel.zip` and `deadline_cloud_for_blender_submitter-deps-macos-arm64.zip`? We should be cautious when doing destructive operations with file globs, there's potential for some pretty bad and unintentional side effects. If possible we should consider deleting specific files.

installer/DeadlineCloudForBlenderSubmitter.xml
```diff
@@ -0,0 +1,288 @@
+<project>
+    <shortName>DeadlineCloudForBlenderSubmitter</shortName>
+    <fullName>AWS Deadline Cloud for Blender Submitter</fullName>
+    <version>1.0</version>
+    <vendor>Amazon Web Services, Inc.</vendor>
+    <!--
+        The installer filename will take one of the following forms:
+        - DeadlineCloudForBlenderSubmitter-windows-x64-installer.exe
+        - DeadlineCloudForBlenderSubmitter-linux-x64-installer.run
+        - DeadlineCloudForBlenderSubmitter-osx-x64-installer.app
```
edwards-aws: Are there different names for intel and arm64 installers?

installer/DeadlineCloudForBlenderSubmitter.xml
```diff
@@ -0,0 +1,288 @@
...
+    <windowsExecutableIcon>icon.ico</windowsExecutableIcon>
+    <windowsUninstallerExecutableIcon>icon.ico</windowsUninstallerExecutableIcon>
+
+    <!-- Allow Windows installer to run as regular user, since running as Admin is forced by default -->
+    <requireInstallationByRootUser>0</requireInstallationByRootUser>
+    <requestedExecutionLevel>asInvoker</requestedExecutionLevel>
+
+    <readyToInstallActionList>
+        <!-- Cleanup Submitters Directory Prior to Installation -->
+        <showQuestion>
```
edwards-aws: What's the behaviour here if someone gets half-way through an installation and cancels? Are the previous files gone forever now?

scripts/build_installer.py
```diff
@@ -0,0 +1,378 @@
...
+    def componentName(self) -> str:
+        """
+        The component name that corresponds to a subdirectory of 'components' subdir
+        of the InstallBuilder project file's directory. By convention, this should
+        match the submitter's repository name.
+        """
+        return f"deadline-cloud-for-{self.name}"
+
+
+class BadRCError(Exception):
```
edwards-aws: Can we get brief docstrings explaining the use case of these exceptions?

scripts/build_installer.py
```diff
@@ -0,0 +1,378 @@
...
+        kwargs["shell"] = False
+    if cwd is not None:
+        kwargs["cwd"] = cwd
+    if env is not None:
+        kwargs["env"] = env
+    p = subprocess.Popen(cmd, **kwargs)
+    stdout, stderr = p.communicate()
+    output = stdout.decode("utf-8") + stderr.decode("utf-8")
+    if p.returncode != 0:
+        raise BadRCError(f"Bad rc ({p.returncode}) for cmd '{cmd}': {output}")
```
edwards-aws: Is `RC` return code? We might want to avoid acronyms here so that the operator can quickly understand the output.

scripts/build_installer.py
```diff
@@ -0,0 +1,378 @@
...
+            install_builder,
+            "build",
+            os.path.join(INSTALLER_ROOT, INSTALLER_TEMPLATE),
+            platform,
+            "--setvars",
+            f"project.outputDirectory={out_dir}",
+            f"project.version={installer_version[:8]}-{date}",
+        ]
+    )
+    sys.stdout.write(
```
edwards-aws: Curious, why `sys.stdout.write` and not `print()`?

scripts/build_installer.py
```diff
@@ -0,0 +1,378 @@
...
+def run(cmd, cwd=None, env=None, echo=True):
+    if echo:
+        sys.stdout.write(f"Running cmd: {cmd}\n")
+    kwargs = {
+        "shell": True,
+        "stdout": subprocess.PIPE,
+        "stderr": subprocess.PIPE,
+    }
+    if isinstance(cmd, list):
+        kwargs["shell"] = False
```
edwards-aws: Is there a reason to allow shell here? There are security considerations when passing in `shell=True`, see https://docs.python.org/3/library/subprocess.html#security-considerations.

The biggest thing is that Python won't do any string wrapping for you, so it's possible to end up with some really weird errors.

scripts/build_installer.py
```diff
@@ -0,0 +1,378 @@
...
+    if cwd is not None:
+        kwargs["cwd"] = cwd
+    if env is not None:
+        kwargs["env"] = env
+    p = subprocess.Popen(cmd, **kwargs)
```
edwards-aws: nit: `subprocess.run()` is preferable imo if it's possible to use https://docs.python.org/3/library/subprocess.html#subprocess.run, the interface is more simple. Additionally if you pass the `text=True` argument you don't have to do the stdout and stderr decoding yourself. I'm pretty sure you can use the `text=True` argument in Popen as well.

test/installer/test_installer.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # WHEN
+    result = subprocess.run([uninstaller_path, "--mode", "unattended"], check=True)
+
+    # THEN
+    assert result.returncode == 0
+
+    # The uninstall process will return before the uninstallation is complete.
+    # If necessary, wait for up to one minute before timing out.
+    for _ in range(6):
```
edwards-aws, line 102: Should there be an assertion here if the uninstall times out? If we're expecting the uninstallation to take up to a minute and it doesn't finish then that might mean the uninstaller is hanging and that's a test failure.

edwards-aws (changes requested):

installer/DeadlineCloudForBlenderSubmitter.xml
miabatta: So this is interesting, in the InstallBuilder component file we specify `all` for allowed platforms but only allow `linux-x64`, `windows-x64`, and `osx` to be passed in as arguments. We'd be fine deleting this comment (and the argument) since you can build for _any_ [supported platform](https://installbuilder.com/installbuilder-specifications) in theory.

But to answer your question, they'd be different but follow the same format.

(Edit: updated the link, it was to the wrong docs :))

installer/DeadlineCloudForBlenderSubmitter.xml
miabatta: Not sure how it works for the shared installer currently, but for the scope of these changes I'm realizing it actually doesn't matter. No other submitters would have to be cleaned out, so we can overwrite the existing Blender files, and [those are saved/restored by rollback](https://releases.installbuilder.com/installbuilder/docs/installbuilder-userguide.html#_rollback) in case of an error.

Thanks for bringing it up!

scripts/build_installer.py
miabatta: Great question. I'll switch to `print` for simplicity

scripts/build_installer.py
miabatta: I believe switching to `subprocess.run` might mean we can get rid of this function in general, as well as the RC error. I'll take a look at refactoring. Thanks!

installer/DeadlineCloudForBlenderSubmitter.xml
edwards-aws: Gotcha, tyty.

depsBundle.sh
miabatta: Yeah, you're right. We'll stick to the specific file names.

test/installer/test_installer.py
miabatta: Yup, thought I included one but I totally forgot. Will be in the next fix

scripts/build_installer.py
evanspearman-a: It's not a big deal, so it's okay to leave it if that's easier.

evanspearman-a (dismissed):

scripts/build_installer.py
```diff
@@ -172,63 +124,67 @@ def build_installer(
...
     # Clone dcc component by copying it into the working directory & allowing the dependency bundle script to be executed
     repo_dir = f"{workdir}/{dcc_component.componentName}"
-    source_folder = Path(os.path.abspath(__file__)).parent.parent
+    source_folder = Path(__file__).absolute().parent.parent
     shutil.copytree(source_folder, repo_dir, dirs_exist_ok=True)
-    bundle_file = Path(os.path.join(repo_dir, "depsBundle.sh"))
+    bundle_file = Path(repo_dir) / "depsBundle.sh"
     bundle_file.chmod(bundle_file.stat().st_mode | stat.S_IEXEC)
-    run(str(bundle_file))
+    subprocess.run(str(bundle_file))
```
edwards-aws: Should we do `check=True` here? Otherwise this can fail silently.

test/installer/test_installer.py
```diff
@@ -0,0 +1,106 @@
...
+        if not installed.exists():
+            break
+        time.sleep(10)
+    assert False, "Uninstall timed out"
```
edwards-aws: I think this will always assert `False`. The `break` will exit the loop and then go to the `assert False` line. But I think the `break` is supposed to indicate a success?

edwards-aws (changes requested):
Thanks Mia! I just had a couple questions.

test/installer/test_installer.py
miabatta: Thanks- both issues fixed!

edwards-aws (approved):
Thanks Mia!

evanspearman-a (approved):

PR #201, opened 2025-03-12
fix(installer): Adapt installer script for prod builds

miabatta:
The existing script to build the Blender installer needs some revisions to work correctly for production builds.

- Removed `install_builder_location` as a universally required argument; instead it is recommended for local builds only
- Replaced the `dev_create_dcc_component` function with copying over required files to `install_builder/components` as necessary
- Edited the pipeline `build_installer` script to take & supply required arguments to `hatch build-installer`
- Other error handling/typing refactors as necessary

We can use this script for production builds!

- Ran the new script for both local builds & in a mock prod environment.
- Confirmed the unit tests still pass.

No change to submitter/adaptor, skipped

Updated docstrings and CLI help, instructions in `DEVELOPMENT.md` have not changed

No

----

### DIFF
--- a/scripts/build_installer.py
+++ b/scripts/build_installer.py
@@ -82,8 +82,8 @@ class EvaluationBuildError(Exception):
     pass
 
 
-def download_from_s3(bucket_name: str, key: Path, output_folder: Path) -> Path:
-    dest_path = output_folder / key.name
+def download_from_s3(bucket_name: str, key: str, output_folder: str) -> Path:
+    dest_path = Path(output_folder) / Path(key).name
     print(f"Downloading {key} from s3:\\\\{bucket_name}")
     import boto3
 
@@ -124,7 +124,7 @@ def build_installer(
     else:
         install_builder_path = install_builder_location
 
-    if not install_builder_path.is_dir():
+    if not Path(install_builder_path).is_dir():
         raise FileNotFoundError(
             f"InstallBuilder path '{str(install_builder_path)}' must be a directory containing 'bin/builder'."
         )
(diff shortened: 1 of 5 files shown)


scripts/build_installer.py
```diff
@@ -217,14 +209,20 @@ def main(args: argparse.Namespace) -> None:
...
-            dev_create_dcc_component(workdir, dcc_submitter)
 
-        src_component_path = f"{workdir}/{dcc_submitter.componentName}"
-        dst_component_path = Path(components_dir) / dcc_submitter.componentName
-        if Path(dst_component_path).exists():
-            shutil.rmtree(dst_component_path, onerror=_add_write_perms)
-        shutil.copytree(src_component_path, dst_component_path)
+        if components_dir.exists():
+            shutil.rmtree(components_dir, onerror=_add_write_perms)
+        src_component_path = Path("./install_builder")
```
epmog: I believe we already have this path with `INSTALL_BUILDER_PROJECT_ROOT`? Removes the need to have a specific CWD.

FWIW we'll be deleting the install_builder folder once we no longer use that component

test/installer/test_installer.py
```diff
@@ -60,11 +60,11 @@ def test_install(installed: Path):
         uninstaller = "uninstall.exe"
     else:
         uninstaller = "uninstall"
-    python_dir = installed / "Submitters" / "Blender" / "python"
+    python_dir = installed / "python"
```
epmog, line 63: I must have glanced over it, what ended up changing this value?

It looks like it should still be the same. Although I do see this variable get overridden... https://github.com/aws-deadline/deadline-cloud-for-blender/pull/201/files#diff-e270a7ac513c8942a2cde57d836ca1c409f2fd6771e29195e6990f6c7f8ecae9L49-L54

miabatta: Now that we're testing individual installs, the Blender files are placed directly in `DeadlineCloudForBlenderSubmitter` instead of creating `Submitters/Blender`. In the code you linked, the install directory changes based on whether the component is being used in a Blender-specific install or a shared install.

scripts/build_installer.py
miabatta: Yep I can replace that, thanks!

.github/workflows/build_installers.yml
```diff
@@ -9,7 +9,7 @@ on:
 
 jobs:
   BuildLinuxInstaller:
-    if: (github.repository == 'aws-deadline/deadline-cloud-for-blender')
+    if: (github.repository == 'miabatta/deadline-cloud-for-blender')
```
epmog: reminder to remove this change before merging

epmog (approved):

erico-aws (approved):

PR #204, opened 2025-03-17
test: add installer permission tests

epmog:

We need automated tests that validate the installer behaves as expected. Default values, file permissions, evaluation mode, signing, etc.

Add them!

We've got automated tests that help ensure we're not breaking anything!

These tests are not currently run automatically in GitHub, so I performed this iteration cycle across the different operating systems
```
hatch run build-installer --platform <osx|...> --local-dev-build --install-builder-location <...>
hatch run test-installer
```

I've ran through these tests a few times on each operating system.

N/A

N/A

N/A

----

### DIFF
--- a/scripts/build_installer.py
+++ b/scripts/build_installer.py
@@ -218,7 +218,7 @@ def main(args: argparse.Namespace) -> None:
         shutil.copy("depsBundle.py", f"{components_dir}/depsBundle.py")
         bundle_file.chmod(bundle_file.stat().st_mode | stat.S_IEXEC)
         try:
-            subprocess.run(str(bundle_file), check=True)
+            subprocess.run(["bash", bundle_file], check=True)
         except subprocess.CalledProcessError as e:
             print(f"Encountered the following error when bundling dependencies: {e.output}")
             raise
--- a/test/installer/test_installer.py
+++ b/test/installer/test_installer.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 import getpass
 import os
 from pathlib import Path
+import re
 import stat
 import subprocess
 import platform
@@ -179,6 +180,55 @@ def uninstaller_path():
         uninstaller_path = uninstaller_path.with_suffix("exe")
 
     yield uninstaller_path
+
+
+def test_default_location(installer_path: Path):
+    """Ensures that the default output location reported by the installer is accurate.
+       The help text will only show it for the default scope (user). Example help output:
+
+    --prefix <prefix>                           Installation Directory
+                                                Default: /home/<user>/DeadlineCloudForBlenderSubmitter
+    """
+    # GIVEN
+    default_install_location = Path("~/DeadlineCloudForBlenderSubmitter").expanduser()
+    default_pattern = r"Default: (.*)"
+    location = ""
+
+    # WHEN
+    help_result = subprocess.run(
+        [installer_path, "--mode", "text", "--help"], check=True, capture_output=True
+    )
+
+    # THEN
+    assert help_result.returncode == 0
+
+    help_output = iter(help_result.stdout.decode("utf-8").splitlines())
+    while (line := next(help_output, None)) is not None:
+        if line.strip().startswith("--prefix"):
+            location = re.match(default_pattern, next(help_output, "").strip(), flags=re.IGNORECASE)
+            break
+
+    assert (
+        location
+    ), f"Could not find default install location in help output:\n{help_result.stdout.decode('utf-8')}"
+    if platform.system() != "Windows":
+        assert location.group(1) == default_install_location.as_posix()
+    else:
+        assert location.group(1) == str(default_install_location())
+
+
+def test_does_not_use_evaluation_mode(installer_path: Path, tmp_path: Path):
+    """Ensure the installer is not built with the evaluation mode of the installer software"""
+    # GIVEN
+    eval_text = r"Created with an evaluation version of InstallBuilder"
+
+    # WHEN
+    result = subprocess.run(
+        [installer_path, "--mode", "text", "--prefix", tmp_path], check=True, capture_output=True
+    )
+
+    # THEN
+    assert eval_text not in result.stdout.decode("utf-8")
 
 
 @pytest.mark.skipif(platform.system() != "Darwin", reason="Only run on macOS")
--- a/test/installer/test_installer.py
+++ b/test/installer/test_installer.py
@@ -1,18 +1,25 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
-from collections import defaultdict
 import getpass
+import glob
 import os
-from pathlib import Path
+import platform
 import re
+import shutil
 import stat
 import subprocess
-import platform
+import sys
+import time
+from collections import defaultdict
+from pathlib import Path
 
 import pytest
 
 
 def _is_admin() -> bool:
-    if platform.system() != "Windows":
+    """Platform independent utility to determine if the tests are running with
+    elevated privileges"""
+    # sys.platform helps mypy type checking ignore other platforms
+    if sys.platform != "win32":
         return os.getuid() == 0
 
     import ctypes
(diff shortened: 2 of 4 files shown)


test/installer/test_installer.py
```diff
@@ -33,45 +55,45 @@ def installer_path():
...
+    top_level_dir = [f.name for f in installation_path.iterdir()]
+    assert "python" in top_level_dir
     assert "installer_version.txt" in top_level_dir
     assert uninstaller in top_level_dir
 
     # Just check that we have dependencies in this folder
     module_dir = [f.name for f in (python_dir / "modules").iterdir()]
     assert "deadline" in module_dir
     assert "qtpy" in module_dir
+    assert "xxhash" in module_dir
```
lucaseck, line 96: I could see some value in checking here that the correct version of xxhash is being included and for the right python version e.g. `_xxhash.cp310-win_amd64`?

test/installer/test_installer.py
```diff
@@ -80,27 +102,436 @@ def test_install(installed: Path):
...
+    uninstaller_path = Path("uninstall")
+    if platform.system() == "Darwin":
+        uninstaller_path = Path("uninstall.app", "Contents", "MacOS", "installbuilder.sh")
+    elif platform.system() == "Windows":
+        uninstaller_path = uninstaller_path.with_suffix(".exe")
+
+    yield uninstaller_path
+
+
+@pytest.mark.xfail(platform.system() == "Windows", reason="'text' mode does not exist on Windows")
```
lucaseck: nit: Why not just skip if it's never going to work?

test/installer/test_installer.py
```diff
@@ -80,27 +102,436 @@ def test_install(installed: Path):
...
+        # GIVEN / WHEN
+        result = subprocess.run(
+            [per_test_user_installation / uninstaller_path, "--mode", "unattended"]
+        )
+
+        # THEN
+        assert result.returncode == 0
+
+        # On Windows, the uninstall process will return before the uninstallation is complete.
+        # If necessary, wait for up to one minute before timing out.
```
lucaseck: nit: 10 * 10 is not one minute

test/installer/test_installer.py
```diff
@@ -80,27 +102,436 @@ def test_install(installed: Path):
...
+        if platform.system() == "Windows":
+            for _ in range(10):
+                if not per_test_system_installation.exists():
+                    break
+                time.sleep(10)
+        assert not per_test_system_installation.exists()
+
+
+@pytest.mark.skipif(
+    os.getenv("CODEBUILD_SRC_DIR") is not None,
```
lucaseck: Might be misunderstanding but don't we only want these to run if `CODEBUILD_SRC_DIR` is set meaning we are running internally? This looks like it's skipping these when on CodeBuild right now.

test/installer/test_installer.py
```diff
@@ -33,45 +55,45 @@ def installer_path():
     yield Path(path).absolute()
 
 
-@pytest.fixture(scope="function")
-def installed(installer_path, tmp_path):
+def _run_installer(installer_path, install_scope, installation_path) -> Path:
+    # use a path that does not exist
+    installation_path = installation_path / "dne"
```
lucaseck, line 60: What's the purpose of this, I thought these were always running in new temp directories to begin with?

epmog: we want the installer to create the path so that it also uninstalls it. If the folder already exists installbuilder may not delete the folder. I encountered this mostly on Windows.

test/installer/test_installer.py
epmog: Cause we're gonna get it to work :)

test/installer/test_installer.py
epmog: Oops! You're correct, fixing

test/installer/test_installer.py
epmog: You're right in that we can do more that type of test may be better suited for the e2e tests that will leverage the installer. In that way there's validation that what is packaged actually runs vs. packaging of what we believe should be what runs. If that makes sense. 

In any case, we've talked offline and we're going to defer this for now

scripts/build_installer.py
```diff
@@ -218,7 +218,7 @@ def main(args: argparse.Namespace) -> None:
         shutil.copy("depsBundle.py", f"{components_dir}/depsBundle.py")
         bundle_file.chmod(bundle_file.stat().st_mode | stat.S_IEXEC)
         try:
-            subprocess.run(str(bundle_file), check=True)
+            subprocess.run(["bash", bundle_file], check=True)
```
lucaseck, line 221: Just confirming, we are making git-bash etc a requirement if running on Windows?

lucaseck (dismissed):

scripts/build_installer.py
epmog: This will change based on deadline-client changes. This was done to unblock what we currently have. Effectively a no-op.

miabatta: We are changing this script regardless after [corresponding changes in `deadline-cloud`](https://github.com/aws-deadline/deadline-cloud/pull/629) are merged. I believe those changes use `shell=True` in an equivalent `run` function which should work on Windows regardless. It has some security considerations but that's a topic for the other PR.

test/installer/test_installer.py
```diff
@@ -80,27 +102,431 @@ def test_install(installed: Path):
...
+        import ntsecuritycon
+        import win32con
+        import win32file
+        import win32security
+
+        # GIVEN
+        windows_user = getpass.getuser()
+        builtin_admin_group_sid, _, _ = win32security.LookupAccountName(None, "Administrators")
+        user_sid, _, _ = win32security.LookupAccountName(None, windows_user)
+        users_group_sid, _, _ = win32security.LookupAccountName(None, "Users")
```
epmog: Noticed that this is unused, I'll remove if another revision is needed.

test/installer/test_installer.py
```diff
@@ -80,27 +102,431 @@ def test_install(installed: Path):
...
+        # GIVEN / WHEN
+        result = subprocess.run(
+            [per_test_user_installation / uninstaller_path, "--mode", "unattended"]
+        )
+
+        # THEN
+        assert result.returncode == 0
+
+        # On Windows, the uninstall process will return before the uninstallation is complete.
+        # If necessary, wait for up to 1 minute 40 seconds before timing out.
```
AWS-Samuel, line 396: Is there a way to detect if the uninstallation is still running (say, query running procs) that we could use instead? Or force the uninstall to be blocking? ~2 minutes is a bit of a long wait time if the uninstallation failed quickly

Applies here and below

test/installer/test_installer.py
AWS-Samuel, line 60: Would a [tempdir](https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir) be better so that the tests could be run in parallel?

AWS-Samuel (dismissed):
Two nits, that aren't really blocking. 

Should we add instructions in the PR template so that contributors know when/how they should run the installer tests?

test/installer/test_installer.py
epmog: This one still has to be root-caused (and existed previously). I'll create an item to tackle this afterwards.

test/installer/test_installer.py
epmog: it IS a tmpdir, from the calling function, but we do currently limit to 1 test at a time. It was my memory that you would be blocked from running more than 1 installer, but that might only be if an option is enabled. I'll give it a whirl!

epmog: It happens to work (I had already properly set-up the test fixtures), nice! I no longer restrict these tests to 1 worker.

AWS-Samuel: Nice, thanks for investigating

AWS-Samuel (approved):

AWS-Samuel (approved):

lucaseck (approved):

PR #213, opened 2025-03-28
test: let windows installer tests run on containers

edwards-aws:
Installer ownership tests were failing on Windows on Containers

The admin group is `ContainerAdministrator` instead of `Administrators` on Windows containers.

Tests can run on Windows Containers, and therefore the pipeline.

Ran the tests in CodeBuild. I haven't had a chance to run on a local Windows machine. Will do that will this is being reviewed.

Doesn't touch any submitter/adaptor code

N/A

N/A

No

----


lucaseck (approved):

miabatta (approved):
Works locally, thanks Cody!

PR #232, opened 2025-05-22
ci: update release workflow

moorec-aws:
There are changes to the [re-usable workflows](https://github.com/aws-deadline/.github/pull/39), release notes are not written and extra validation required for tag creation at release.

Depends on these changes: https://github.com/aws-deadline/.github/pull/39

1. Updated `code_quality.yml` to accept tags as inputs, and moved the Unit Tests from the `release_bump.yml` to the `release_publish.yml` workflow.
2. The `release_publish.yml` workflow automatically starts after a ChangeLog PR has been merged. We used the new `reusable_tag_release.yml` workflow to create a new tag. We validate the tag has not been released and the author of the associated commit. The workflow can be re-run, specifying an existing tag.
3. Building and Staging the installer has been moved to its own workflow.
4. IsCondaReady was moved to run before we release to GitHub.

More modular workflows, added automation.

Tested the changes in a [fork](https://github.com/moorec-aws/deadline-cloud-for-blender/actions/runs/15218951005), publishing steps were skipped.

No

No

----


miabatta (approved):

AWS-Samuel (approved):

.github/workflows/release_publish.yml
```diff
@@ -1,68 +1,93 @@
...
-          else
-            echo "isReleased=false" >> $GITHUB_OUTPUT
-          fi
-    env:
-      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  TagRelease:
+    uses: aws-deadline/.github/.github/workflows/reusable_tag_release.yml@mainline
+    secrets: inherit
+    with:
+      tag: ${{ github.event_name == 'workflow_dispatch' && inputs.tag || '' }}
```
erico-aws, line 28: What do you think about having 2 separate workflows. One that starts on push to changelog and another that can be manually started from a tag. The changelog one can run this job and then the rest, where as the manual one can just run the rest. That way we don't have these jobs that have different modes based on how the workflow was started.

moorec-aws: I think having two workflow adds more usage complexity than offering a re-run option to the release workflow. The ideal and common situation will be that this just automatically runs and works. The optional re-run should be the exception.

PR #250, opened 2025-07-18
fix: Disable all components by default so cli installs work as expected.

erico-aws:

Enabling components by default leads to strange behaviour with the cli, specifically the shared installer. Giving the Blender installer a list of enabled components doesn't override the default selections. 

For example, doing `--enabled-components blender_42` will result in all version of Blender being installed.

Change this to be more intuitive by defaulting all components to be unselected.

Running the installer through the CLI will be more intuitive. 

Built locally and tested against local install of Blender.

No

No
----

### DIFF
--- a/install_builder/deadline-cloud-for-blender.xml
+++ b/install_builder/deadline-cloud-for-blender.xml
@@ -101,7 +101,7 @@
         <component>
             <name>blender_36</name>
             <description>Blender 3.6</description>
-            <selected>1</selected>
+            <selected>0</selected>
             <parameterList>
                 <fileParameter>
                     <name>blender_36_path</name>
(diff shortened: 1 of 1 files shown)


miabatta (approved):
shipit

AWS-Samuel (approved):
looks good, but can we change the commit title (or modify when you merge it) to describe the bug being fixed?

In this case I think `fix: installing with --enabled-components flag would always install all blender versions`

erico-aws:
> looks good, but can we change the commit title (or modify when you merge it) to describe the bug being fixed?
> 
> In this case I think `fix: installing with --enabled-components flag would always install all blender versions`

Yeah, I'll use that when I merge.

PR #255, opened 2025-08-04
feat: add up to Blender 4.5 support to installer and repository

lucaseck:
Blender 4.5 released as the new LTS version and we should support it.

Add support for up to Blender 4.5
* Updated READMEs
* Updated installer components
* Fixed a bug where spaces in the Windows path name were causing failures when trying to run the plugin script to install within Blender
* Removed some unnecessary variable setting for the Mac installers on uninstallation(this was verified that it works without even in non-default install locations)

The submitter now reflects that it works with Blender 4.5 and in the near future it will be available as an option to install for in the submitter installer

* Manually built installers for Windows, MacOS and Linux
  * Ran the installer on each OS with Blender 4.5 selected and verified it was installed/uninstalled correctly
  * Changed the install path of Blender to be non-default and made sure things still installed/uninstalled correctly
 
image

The existing tests all passed

There are docs updates as part of this

No

----

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -17,7 +17,7 @@ hatch run build
 
 ### Build the installer
 ```bash
-hatch run installer:build-installer --local-dev --platform <PLATFORM> [--install-builder-location <LOCATION> --output-dir <DIR>]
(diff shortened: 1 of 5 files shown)


erico-aws (approved):
Can you add a screenshot of the new components?

miabatta (approved):
Thanks for cleaning up the instructions as well!

PR #257, opened 2025-08-05
fix: unnecessary brush assets included as job attachments

lucaseck:
In versions of Blender 4.3+ even the default cube scene was detecting additional assets to upload. These assets appear to be built-in brush assets in the Blender install. Even if a brush hasn't been used yet in the scene at least one seems to be returned by the call to `bpy.utils.blend_paths()`. It's unclear if this is intentional behaviour or if `bpy.utils.blend_paths` should filter these out since the [API docs](https://docs.blender.org/api/current/bpy.utils.html#bpy.utils.blend_paths) say that function returns __external__ files only(which I guess these are since they are external to the scene).
Based on my reading this was a [change](https://code.blender.org/2024/07/brush-assets-is-out/#new-brush-workflow) in Blender 4.3+ where brushes are no longer baked into the .blend file but instead stored as external assets, including the built-in ones. My understanding of [Blender brushes](https://docs.blender.org/manual/en/latest/sculpt_paint/brush/introduction.html) is that they are used to manipulate the current scene by modifying the geometry or visuals and you can save out different brushes with different characteristics. I couldn't find anything suggesting that these brush assets are actually needed at render time once the scene is saved and sent to the farm. I also did not run into any issues when submitting renders without including the brush assets that were being returned by `bpy.utils.blend_paths`.

Therefore, I've added an additional filter check where we collect the list of input attachments and filter out any that are located within the local Blender resources directory(from calling bpy.utils.resource_paths("LOCAL")) and start with `essentials_brushes-`.

No seemingly unnecessary files being detected as job attachments.

* Created a new unit test to cover the find_files logic
* Ran manual submission tests on Windows in Blender 4.5 and 3.6.
  * Verified that the brush asset was being returned by the `bpy.utils.blend_paths()` call in 4.5, but not being added as an attachment. 
  * Verified that the jobs ran on both 3.6 and 4.5 and no notable errors in the logs due to the missing brush asset on 4.5. Outputs also looked correct.
```
test/integ/test_blender_adaptors.py::TestAdaptors::test_minimal_scene_adaptor
[gw0] [ 25%] PASSED test/integ/test_blender_adaptors.py::TestAdaptors::test_minimal_scene_adaptor
test/integ/test_blender_adaptors.py::TestAdaptors::test_minimal_scene_with_host_requirements_adaptor
[gw0] [ 50%] PASSED test/integ/test_blender_adaptors.py::TestAdaptors::test_minimal_scene_with_host_requirements_adaptor
test/integ/test_blender_submitters.py::TestSubmitters::test_minimal_scene_submitter
[gw0] [ 75%] PASSED test/integ/test_blender_submitters.py::TestSubmitters::test_minimal_scene_submitter
test/integ/test_blender_submitters.py::TestSubmitters::test_minimal_scene_with_host_requirements_submitter
[gw0] [100%] PASSED test/integ/test_blender_submitters.py::TestSubmitters::test_minimal_scene_with_host_requirements_submitter

============================================================================================== slowest 5 durations ==============================================================================================
135.04s call     test/integ/test_blender_adaptors.py::TestAdaptors::test_minimal_scene_adaptor
62.24s call     test/integ/test_blender_adaptors.py::TestAdaptors::test_minimal_scene_with_host_requirements_adaptor
9.14s call     test/integ/test_blender_submitters.py::TestSubmitters::test_minimal_scene_with_host_requirements_submitter
8.27s call     test/integ/test_blender_submitters.py::TestSubmitters::test_minimal_scene_submitter
0.01s setup    test/integ/test_blender_adaptors.py::TestAdaptors::test_minimal_scene_adaptor
========================================================================================= 4 passed in 217.80s (0:03:37) =========================================================================================
```
No

No, it shouldn't be. The only case where I could see an argument for it being possibly breaking is if a user was already for some reason creating custom asset files inside __Blender's__ resource folder with the same name prefix as the internal ones.

----

### DIFF
--- a/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/blender_utils.py
+++ b/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/blender_utils.py
@@ -78,20 +78,42 @@ def find_files(project_path, skip_temp=True, skip_nonexistent=True) -> list[Path
     files.append(project_path)
     files = set(Path(f) for f in files)
 
+    temp_dirs = []
     if skip_temp:
         temp_dirs = _get_blender_temp_dirs()
         _logger.debug(f"Resolved Blender temp directories: {temp_dirs}")
 
-        def is_in_temp(f: Path):
-            """Returns True if the given file is in any of Blender's temp directories."""
-            return any(f.is_relative_to(temp_dir) for temp_dir in temp_dirs)
+    # Path where Blender stores its built-in brush asset .blend files
+    blender_resource_path = Path(bpy.utils.resource_path("LOCAL"))
 
-        files = (f for f in files if not is_in_temp(f))
+    def _is_in_temp(f: Path) -> bool:
+        """Returns True if the given file is in any of Blender's temp directories."""
+        return any(f.is_relative_to(temp_dir) for temp_dir in temp_dirs)
 
-    if skip_nonexistent:
-        files = (f for f in files if f.exists())
+    def _is_essential_brush(path: Path) -> bool:
+        """
+        Returns True if the given file is a built-in brush asset.
+        Any paths to files within the local Blender resource folder prefixed with
+        'essentials_brushes-' are filtered out since these are bundled with Blender and appear
+        to be unneccesary for just rendering. These files were previously included in the main
+        .blend file but with 4.3+ they are now separate assets in the Blender install
+        https://code.blender.org/2024/07/brush-assets-is-out/#new-brush-workflow
+        """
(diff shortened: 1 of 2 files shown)


epmog (approved):

miabatta (changes requested):
Can we change the commit message so we don't have to retroactively fix it in the changelog?

e.g., `fix: Built-in brush assets are included as job attachments`

miabatta (approved):

miabatta:
Awesome, thanks!

PR #290, opened 2025-10-15
docs: Create Blender submitter user guide

miabatta:
We are lacking user-facing documentation for the submitter. Only developers like READMEs :'(

Created a `mkdocs` site with Blender submitter documentation to be consumed & hosted on the shared integrations page.

Users will have a better starting point for getting help!

Served the page on localhost and tried it out.

Demo video here: 

https://github.com/user-attachments/assets/51ab40c7-60ae-4c63-9e37-2e56bcec05c2

n/a, no submitter/adaptor changes

n/a

this change _is_ documentation

nop

----


docs/user_guide/images/job-specific-settings.png
lucaseck, line 1: The changes to use the new login/logout UX are merged and will be in the next release. Thoughts on updating the screenshots now so we don't forget to later?

docs/user_guide/installation/index.md
```diff
@@ -0,0 +1,15 @@
+# Installing AWS Deadline Cloud for Blender 
+
+To install the AWS Deadline Cloud for Blender submitter, you will need:
+
+- A Windows, MacOS, or Linux workstation
+- Blender 3.6 or later
+- [Deadline Cloud monitor](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/monitor-onboarding.html)
+- Access to an AWS Deadline Cloud farm with either
+    - a service-managed fleet, or
```
lucaseck: nit: Nothing on SMF requires users to use our Conda packages that provide the adaptors. If we are calling out the adaptor for CMF is it worth calling out the same things, but mentioning that the default Conda settings will handle it all by default for SMF?

docs/user_guide/installation/index.md
```diff
@@ -0,0 +1,15 @@
...
+    - a customer-managed fleet with Blender and the Blender adaptor set up
+
+There are two ways to install the Deadline Cloud for Blender submitter:
```
lucaseck: Thoughts on including the third manual option here from the DEVELOPMENT.md? I personally feel like it's not super obvious that a manual option exists and is a decent last resort if all of the other methods are failing for some reason for a user. It's maybe more developer focussed than needed for a "user guide".

docs/user_guide/installation/installation.md
```diff
@@ -0,0 +1,31 @@
...
+You can install the Deadline Cloud for Blender submitter using the Deadline Cloud submitter installer.
+
+**To install the submitter**
+
+1. Download the [Deadline Cloud submitter installer](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html).
+1. Run the installer and follow the prompts.
+1. Launch Blender.
+
+The Deadline Cloud submitter add-on should be automatically enabled.
+> The submitter installer is available for Windows, MacOS, and Linux. See the [developer README](https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/README.md) for manual installation instructions.
```
lucaseck: I think this link might be wrong, the manual installation instructions are in the DEVELOPMENT.md.

nit: should we call our specifically ARM for MacOS since that confusion has come up before.

docs/user_guide/installation/installation.md
```diff
@@ -0,0 +1,31 @@
...
+If the add-on is not available from the **Render** menu, you will need to manually enable it.
+
+**To manually enable the submitter add-on**
+
+1. On the **Edit** menu, choose **Preferences…**.
+1. Choose **File Paths** on the left side bar.
+1. Find the **Script Directories** section and choose **+**.
+1. For **Name**, enter `python`.
+1. For **Path**, enter the path to the `python` directory in your Blender submitter installation.
+1. Restart Blender for changes to take effect.
```
lucaseck, line 22: I don't think we have any "common" installation issues right now, but if there are any debugging steps for users to follow if the installation didn't work for them can we put them here. e.g. where to look for errors in the console

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,38 @@
+# Using the Deadline Cloud for Blender submitter
+
+## Submit a job
+
+**To submit a job from Blender to Deadline Cloud**
+
+1. Save your Blender file.
+1. In Blender's **Render** menu, choose **Submit to AWS Deadline Cloud**.
+    - You may see a pop-up to install GUI dependencies. Choose **OK** and the submitter dialog will appear shortly.
+1. Choose **Login** at the bottom of the dialog to log into your AWS profile.
```
lucaseck: Login is only an option if using DCM provided credentials which isn't a hard requirement to submit

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,38 @@
...
+1. (Optional) To export a job bundle to your job history directory without submitting it, choose **Export bundle**.
+1. Choose **Submit** and follow the prompts to send your job to Deadline Cloud.
+
+## Blender-specific Settings
+The **Job-specific settings** tab has options specific to jobs created in Blender.
+![Submitter Interface](./images/job-specific-settings.png)
+
+  - *Project Path* - The location where the current project is saved. Can't be changed.
+  - *Output Directory* - The location to save file outputs from the render job.
+  - *Output File Prefix* - The pattern to use when naming file outputs. Follows Blender's convention for file names.
```
lucaseck: nit: should we call out the current "issue" here where the render layer is always appended to the output and that's why this is just the prefix?

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,38 @@
...
+  - *Scene* - The scene from the current project to render.
+  - *Render Engine* - The render engine (Cycles, EEVEE, or Workbench) to use.
+  - *View Layers* - The layer to render, or "All Renderable Layers" to render each applicable layer in the scene separately.
+  - *Cameras* - The camera to render, or "All Renderable Cameras" to render each camera in the scene separately.
+  - *Cycles GPU Rendering* - Select this to enable GPU rendering. Choose a device type supported by Blender or specify your own.
```
lucaseck: Is there a possible "gotcha" here where the user needs to be aware of what device types are supported on their render fleets? We don't do any pre-filtering so I'd assume there are some options that just won't work on the `g` instances that are available in SMF and might only be applicable to CMF

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,38 @@
...
+  - *Override Frame Range* - Select this to render a different frame or frame range than is specified in the scene file.
```
lucaseck: Do we have any docs in openjd to link to on the formatting accepted here for the frame range?

lucaseck (commented):
Some small suggestions and questions, but otherwise looks awesome! Thanks for taking the time to improve the documentation 🥳

docs/user_guide/images/job-specific-settings.png
miabatta: Great callout, I will add some screenshots using `deadline-cloud` mainline.
Part of the doc writing guidance will be screenshot maintenance so if we expect many more changes, it's also a reasonable tradeoff to exclude them. But we can decide that fairly quickly later on :)

docs/user_guide/installation/index.md
crowecawcaw: We mention that it exists and link to the DEVELOPMENT.md file so someone can find it if they're looking here for it but without duplicating the content or adding intimidating CLI instructions.

docs/user_guide/installation/index.md
crowecawcaw: Are you pointing out that this might be more correct as, "a service-managed fleet and default queue environment"? I think adding too much detail here could make the instructions harder to understand for an end users. Is there a short hand way of saying what's required? Or maybe we could link to the GitHub README for detailed setup instructions?

docs/user_guide/installation/installation.md
```diff
@@ -0,0 +1,31 @@
...
+
+You can install the Deadline Cloud for Blender submitter using the Deadline Cloud submitter installer.
+
+**To install the submitter**
+
+1. Download the [Deadline Cloud submitter installer](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html).
+1. Run the installer and follow the prompts.
+1. Launch Blender.
+
+The Deadline Cloud submitter add-on should be automatically enabled.
```
crowecawcaw: Is this new? It previously required a manual step.

docs/user_guide/installation/installation.md
```diff
@@ -0,0 +1,31 @@
...
+> The submitter installer is available for Windows, MacOS, and Linux. See the [developer README](https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/README.md) for manual installation instructions.
+
+**To verify the submitter is installed correctly**
```
crowecawcaw: Can someone verify it's installed by looking for the option in the Render menu? These steps feel a bit more like debugging steps. Could be renamed something like, "Troubleshooting".

docs/user_guide/telemetry.md
```diff
@@ -0,0 +1,23 @@
...
+This library collects telemetry data by default. Telemetry events contain non-personally identifiable information that helps us understand how users interact with our software. We use this to know which features our customers use and/or what existing pain points are.
+
+You can opt out of telemetry on your machine by editing your Deadline Cloud configuration, or temporarily by setting an environment variable in your terminal.
+
+**To opt out of telemetry (config)**
+
+1. Run the `deadline config` command:
+```sh
+deadline config set telemetry.opt_out true
+```
```
crowecawcaw: Telemetry opt out is also a checkbox in the submitter settings page I think. What about suggesting the GUI method instead here?

docs/user_guide/using-submitter.md
crowecawcaw: Maybe something like, "If you see a Login button, click it to..."

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,38 @@
...
+## Submit a job
+
+**To submit a job from Blender to Deadline Cloud**
+
+1. Save your Blender file.
+1. In Blender's **Render** menu, choose **Submit to AWS Deadline Cloud**.
+    - You may see a pop-up to install GUI dependencies. Choose **OK** and the submitter dialog will appear shortly.
+1. Choose **Login** at the bottom of the dialog to log into your AWS profile.
+1. Use the tabs in the dialog to customize your job.
+1. (Optional) To export a job bundle to your job history directory without submitting it, choose **Export bundle**.
```
crowecawcaw: I think this could either use some explanation (what's a bundle and why might I want to export it?) or could be dropped.

docs/user_guide/using-submitter.md
crowecawcaw: Also it sounds like the frame range is usually pulled from the scene file. Could say that specifically and with a note like, "Leave this field blank to use the frame range from your Blender scene."

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,38 @@
...
+  - *Output Directory* - The location to save file outputs from the render job.
+  - *Output File Prefix* - The pattern to use when naming file outputs. Follows Blender's convention for file names.
+  - *Scene* - The scene from the current project to render.
+  - *Render Engine* - The render engine (Cycles, EEVEE, or Workbench) to use.
+  - *View Layers* - The layer to render, or "All Renderable Layers" to render each applicable layer in the scene separately.
+  - *Cameras* - The camera to render, or "All Renderable Cameras" to render each camera in the scene separately.
+  - *Cycles GPU Rendering* - Select this to enable GPU rendering. Choose a device type supported by Blender or specify your own.
+  - *Override Frame Range* - Select this to render a different frame or frame range than is specified in the scene file.
+
+For information about the other submitter tabs, see the [AWS Deadline Cloud guide for using a submitter](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/jobs-using-submitter.html).
```
crowecawcaw, line 34: Nice!

docs/user_guide/using-submitter.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Monitoring your jobs
+
+You can monitor job progress using the Deadline Cloud monitor. For more information, see the [AWS Deadline Cloud guide for using the monitor](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/working-with-deadline-monitor.html).
+
+## Getting help
+
+- Contact AWS Support
+- (Requires a GitHub account) [Open an issue in `deadline-cloud-for-blender` on GitHub](https://github.com/aws-deadline/deadline-cloud-for-blender/issues)
```
crowecawcaw, line 43: This would make a great common page later.

docs/user_guide/installation/index.md
miabatta: I am a little mixed on this because I want to avoid overloading new users with info. In the interest of completeness I could add a note saying that you can set up the adaptor on SMF in non-conda ways, but the intent of this phrasing is to reflect that you _can_ submit to an SMF with no further action.

Hope that makes sense. I will still add some supplementary info on that though!

docs/user_guide/installation/index.md
miabatta: Sounds good, the instructions are there so no harm in linking

docs/user_guide/installation/installation.md
miabatta: For sure! If we can collect FAQs over time it would be great to have a troubleshooting section in this file

docs/user_guide/using-submitter.md
miabatta: Yep will do. IIRC we've agreed to keep an eye out on whether customers feel strongly about the name reformatting and revisit it if so

docs/user_guide/using-submitter.md
miabatta: Yes, I will link that in this section as well! It would be the Open JD specifications wiki

docs/user_guide/using-submitter.md
miabatta: Not sure the best way of communicating this, but if the GPU option is checked and an incompatible device for the fleet is specified, the adaptor will cycle through all device options to find a compatible one before switching to CPU. So for example, a user who has selected Metal but is using a Linux fleet will be able to use CUDA.

Maybe a short sentence here saying that if a device is unavailable, the adaptor will attempt to fall back to an available device type?

docs/user_guide/installation/installation.md
miabatta: Yes! We added prompts to the installer to pre-enable the extension for each version of Blender the user selects.

Instead of a vague "follow the prompts" I can add an instruction to check off all versions of Blender that you want to install, etc.

docs/user_guide/installation/installation.md
miabatta: Yeah you're right. Instead of a separate "verify" procedure I can add a step to check the menu

docs/user_guide/telemetry.md
miabatta: Yes! Not sure why I missed that in the settings dialog. I am fine with only mentioning the GUI method here

docs/user_guide/installation/index.md
lucaseck: I'm good with whatever here, left it just as a nit because the way I read it made it sound like I only needed the adaptor for CMF. I agree too much detail isn't great either, so only worth it if this will actually trip someone up.

docs/user_guide/using-submitter.md
lucaseck: That behaviour is a bit concerning I think at least in my mind if I'm understanding right. If I select Metal, but try to render on an instance that doesn't support it my job will still run and cost me additional money but not actually render with what I selected? I don't know a ton about the differences between the acceleration options and if they can result in any image differences etc, but unless they are identical that seems like an odd choice to m.

docs/user_guide/installation/index.md
miabatta: Haha, re-examining the wording here neither are _technically_ prerequisites for installing the Blender submitter- just for using it.

But on a more serious note, I'm in favour of keeping it vague. The target audience here may not be in charge of setting up their farm infrastructure. I think in the edits I'll leave it at "a farm and queue"

docs/user_guide/using-submitter.md
miabatta: Might just omit this part. If customers have difficulty with the authentication stuff we can re-add it

docs/user_guide/using-submitter.md
miabatta: Yeah, not in scope for this PR but I see what you mean. The PR where this was added is here: https://github.com/aws-deadline/deadline-cloud-for-blender/pull/159

I will document the behaviour for now and if we decide to change it, we can edit the docs correspondingly. We can chat with others about that offline.

docs/user_guide/installation/installation.md
miabatta: Chatted a bit offline- we will not include terminal or console errors here since the audience may not know much about debugging them. If we change our minds always on the table to add in!

lucaseck (approved):
🚢

docs/user_guide/installation/installation.md
crowecawcaw: Auto-enabling is a great improvement, cool!

crowecawcaw (approved):

PR #293, opened 2025-10-30
test: fix adaptor integration test errors when using Blender 4.5

jusiskin:

The integration tests do not work when testing with Blender 4.5. The images render differently than the expected images that were originally rendered from Blender 4.2.

1.  Add a fixture to detect the blender version
2.  Create version-specific directories of expected images since the renders will differ between Blender versions.
3. The existing expected images were moved under the `4.2` subdirectory
4. New expected images were generated by opening the scene in Blender 4.5 and rendering animations for each layer and camera combination manually using the same filename convention as 4.2.

Blender integration tests will work with 4.2 and 4.5

1.  Installed two copied of Blender (4.2 and 4.5) on MacOS
2. Run the integration tests with `BLENDER_EXECUTABLE` pointing to both 4.2 and 4.5 and confirm the tests pass for both

No, the test code is the documentation

No

----


edwards-aws (dismissed):

test/integ/helpers/image_comparison.py
```diff
@@ -7,18 +7,36 @@
...
         expected = np.asarray(PIL.Image.open(image))
 
         # Check that the two images are the same within a tolerance.
         # It's normal for there to be noise in an output image, so it is unlikely that two
         # renders will be exactly the same.
-        assert np.allclose(actual, expected, atol=2)
+        if not np.allclose(actual, expected, atol=2):
+            # For debugging: Uncomment to write the diff image along-side the expected image with _diff suffix
+            diff = actual - expected
+            PIL.Image.fromarray(diff).save(expected_image_directory / f"{image.name}_diff.png")
```
epmog: based on the comment, should this be commented out?

edwards-aws: I kinda think we should always have this. It'll only happen if the tests fail and it'll usually be useful when the tests fail.

edwards-aws (dismissed):

test/integ/helpers/image_comparison.py
jusiskin: Good catch. I commented this back out

epmog (approved):

edwards-aws (approved):

PR #296, opened 2025-11-17
ci: Add test runner setup script 

moorec-aws:
We want to be able to run integration tests for blender across multiple versions in our CI pipeline

1. Add a new hatch env called `integ-ci` environment. Add a matrix of blender versions. Add two scripts `setup` - responsible for setting up the blender application environments, and `test` - runs the integration test using the matrix versions.

2. Added a script called `setup-runner.py`. It is responsible for setting up a runner with blender application install and any require dependencies. By default it looks for a bucket assigned to the env var `DEPENDENCY_BUCKET` to obtain the installers. If passed `--public-urls` it will grab installers the blender downloads page for the versions specified in the script and match them to a sha256.  If using a linux environment you will want to pass `--install_x11`  to install required deps.

3. Various test environments use temp locations. We added some logic to the blender_utils.py that allows the specification of paths to exclude from blender tmp path filtering.

4. When testing blender cross platform each platform is not guaranteed to render output exactly the same. The atol of 2 was too strict. It will now measure the maximum difference across all pixels and if that difference is > 2 for over 0.5% it will raise an error. This should allow for a bit more wiggle room for cross platform variations but also stay strict enough that we will catch any visual differences.

5. Updated the adaptor tests that are running multiple steps to run per step due to a cleanup race condition in openjd runtime.

CI/CD can run the blender integration tests for 4.2, 4.5

Tested in a developer environment. Confirmed that the tests for Mac, Linux and Windows all passed.

https://github.com/moorec-aws/deadline-cloud-for-blender/actions/runs/19515494134

Not yet

No

----

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -9,7 +9,7 @@ jobs:
     permissions:
       id-token: write
       contents: read
-    uses: moorec-aws/.github/.github/workflows/reusable_integration_test.yml@mainline
+    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
     secrets: inherit
     with:
       repository: ${{ github.event.repository.name }}
@@ -21,7 +21,7 @@ jobs:
     permissions:
       id-token: write
       contents: read
-    uses: moorec-aws/.github/.github/workflows/reusable_integration_test.yml@mainline
+    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
     secrets: inherit
     with:
       repository: ${{ github.event.repository.name }}
@@ -33,7 +33,7 @@ jobs:
     permissions:
       id-token: write
       contents: read
-    uses: moorec-aws/.github/.github/workflows/reusable_integration_test.yml@mainline
+    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
     secrets: inherit
     with:
       repository: ${{ github.event.repository.name }}
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -9,7 +9,7 @@ jobs:
     permissions:
       id-token: write
       contents: read
-    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
+    uses: moorec-aws/.github/.github/workflows/reusable_integration_test.yml@mainline
     secrets: inherit
     with:
       repository: ${{ github.event.repository.name }}
@@ -21,7 +21,7 @@ jobs:
     permissions:
       id-token: write
       contents: read
-    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
+    uses: moorec-aws/.github/.github/workflows/reusable_integration_test.yml@mainline
     secrets: inherit
     with:
       repository: ${{ github.event.repository.name }}
@@ -33,7 +33,7 @@ jobs:
     permissions:
       id-token: write
       contents: read
-    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
+    uses: moorec-aws/.github/.github/workflows/reusable_integration_test.yml@mainline
     secrets: inherit
     with:
       repository: ${{ github.event.repository.name }}
--- a/hatch.toml
+++ b/hatch.toml
@@ -38,8 +38,7 @@ build = "hatch build"
 
 [envs.integ]
 pre-install-commands = [
-  "pip install -r requirements-integ-testing.txt",
-  "pip install -r requirements-testing.txt"
+  "pip install -r requirements-integ-testing.txt -r requirements-testing.txt"
 ]
 
 [envs.integ.scripts]
@@ -49,18 +48,22 @@ test_adaptors = "pytest --no-cov {args:test/integ} -vvv --numprocesses=1 -m adap
 
 [envs.integ-ci]
 pre-install-commands = [
-  "pip install -r requirements-integ-testing.txt",
-  "pip install -r requirements-testing.txt"
+  "pip install -r requirements-integ-testing.txt -r requirements-testing.txt"
 ]
 
 [[envs.integ-ci.matrix]]
-blender = ["4.5.4", "4.2.12"]
+blender = ["4.5.4"]
+python = ["3.11"]
+
+[[envs.integ-ci.matrix]]
+blender = ["4.2.12"]
+python = ["3.11"]
 
 [envs.integ-ci.env-vars]
 BLENDER_VERSION = "{matrix:blender}"
 
 [envs.integ-ci.scripts]
-setup = "python ./pipeline/setup-runner.py --versions {matrix:blender} {args}"
+setup = "python ./pipeline/setup-runner.py --versions {matrix:blender} --python-version {matrix:python} {args}"
 test = "pytest --no-cov {args:test/integ} -vvv --numprocesses=1"
 
 [envs.codebuild.env-vars]
--- a/hatch.toml
+++ b/hatch.toml
@@ -47,6 +47,22 @@ test = "pytest --no-cov {args:test/integ} -vvv --numprocesses=1"
 test_submitters = "pytest --no-cov {args:test/integ} -vvv --numprocesses=1 -m submitter"
 test_adaptors = "pytest --no-cov {args:test/integ} -vvv --numprocesses=1 -m adaptor"
 
+[envs.integ-ci]
+pre-install-commands = [
+  "pip install -r requirements-integ-testing.txt",
+  "pip install -r requirements-testing.txt"
+]
+
+[[envs.integ-ci.matrix]]
+blender = ["4.5.4", "4.2.12"]
+
+[envs.integ-ci.env-vars]
+BLENDER_VERSION = "{matrix:blender}"
+
+[envs.integ-ci.scripts]
+setup = "python ./pipeline/setup-runner.py --versions {matrix:blender} {args}"
+test = "pytest --no-cov {args:test/integ} -vvv --numprocesses=1"
+
 [envs.codebuild.env-vars]
 SKIP_BOOTSTRAP_TEST_RESOURCES="True"
 
--- a/pipeline/setup-runner.py
+++ b/pipeline/setup-runner.py
@@ -120,7 +120,8 @@ def setup_linux(python_version, install_x11=False):
             ]
         )
         if run(["pgrep", "Xvfb"], check=False).returncode != 0:
-            run(["Xvfb", ":99", "-screen", "0", "1024x768x24"], check=False)
+            # Start Xvfb in background
+            subprocess.Popen(["Xvfb", ":99", "-screen", "0", "1024x768x24"])
             run(["sleep", "2"])
         print("\nXvfb started. Run this before tests:")
         print("  export DISPLAY=:99")
--- a/pipeline/setup-runner.py
+++ b/pipeline/setup-runner.py
@@ -12,6 +12,10 @@ import time
 from pathlib import Path
 
 BLENDER_VERSIONS = ["4.2.12", "4.5.4"]
+BLENDER_PYTHON_VERSIONS = {
+    "4.2.12": "3.11",
+    "4.5.4": "3.11",
+}
 USE_PUBLIC_URLS = False
 
 # SHA256 checksums from https://www.blender.org/download/
@@ -64,16 +68,13 @@ def verify_checksum(file_path, expected_checksum):
     return True
 
 
-def setup_linux(install_x11=False):
+def setup_linux(python_version, install_x11=False):
     pkg_mgr = (
         "dnf"
         if subprocess.run("command -v dnf", shell=True, capture_output=True).returncode == 0
         else "yum"
     )
     run(f"{pkg_mgr} update -y", shell=True)
-
-    run("pip install --upgrade pip hatch", shell=True)
-    run("pip install --upgrade -r requirements-development.txt", shell=True)
 
     if install_x11:
         run(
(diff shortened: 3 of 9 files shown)


test/integ/conftest.py
```diff
@@ -12,6 +12,20 @@
...
+    if os.name == "nt":
+        os.environ["BLENDER_EXECUTABLE"] = (
+            f"C:\\Tools\\blender-{blender_version_env}-windows-x64\\blender.exe"
+        )
+    elif sys.platform == "darwin":
+        os.environ["BLENDER_EXECUTABLE"] = (
+            f"/Applications/Blender-{blender_version_env}.app/Contents/MacOS/Blender"
+        )
+    else:  # Linux
+        os.environ["BLENDER_EXECUTABLE"] = f"/opt/blender-{blender_version_env}-linux-x64/blender"
```
jusiskin: Two things:

1.  maybe we should document this new behavior in the integration test section of `DEVELOMENT.md`. What do you think?
2.  we should have an assertion in the `blender_version` fixture to ensure the `BLENDER_VERSION` env var aligns with the output of `blender --version`

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/blender_utils.py
```diff
@@ -74,6 +74,13 @@ def find_files(project_path, skip_temp=True, skip_nonexistent=True) -> list[Path
         skip_temp: if True, skip all files from any of Blender's potential temp directories.
         skip_nonexistent: if True, skip all files that do not exist. When files are shared across machines, Blender may retain memory of original paths; this ensures that all retrieved paths exist on the local filesystem.
     """
+    import os
+
+    # Get allowlist from environment variable (semicolon-separated on Windows, colon-separated on Unix)
+    allowlist_env = os.environ.get("BLENDER_TEMP_ALLOWLIST", "")
+    separator = ";" if os.name == "nt" else ":"
+    temp_allowlist = [Path(p.strip()) for p in allowlist_env.split(separator) if p.strip()]
```
jusiskin: nit: we can use [`os.pathsep`](https://docs.python.org/3/library/os.html#os.pathsep) here

```suggestion
    temp_allowlist = [Path(p.strip()) for p in allowlist_env.split(os.pathsep) if p.strip()]
```

hatch.toml
```diff
@@ -47,6 +47,22 @@ test = "pytest --no-cov {args:test/integ} -vvv --numprocesses=1"
 test_submitters = "pytest --no-cov {args:test/integ} -vvv --numprocesses=1 -m submitter"
 test_adaptors = "pytest --no-cov {args:test/integ} -vvv --numprocesses=1 -m adaptor"
 
+[envs.integ-ci]
+pre-install-commands = [
+  "pip install -r requirements-integ-testing.txt",
+  "pip install -r requirements-testing.txt"
```
jusiskin: We should to combine these into a single `pip install` so that the `pip` solver can satisfy both closures in a single pass. The [docs](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-r) say:

>   This option can be used multiple times.

So, we would have:

```suggestion
  "pip install -r requirements-integ-testing.txt -r requirements-testing.txt"
```

pipeline/setup-runner.py
```diff
@@ -0,0 +1,319 @@
...
+    USE_PUBLIC_URLS = args.public_urls
+    if args.versions:
+        BLENDER_VERSIONS = args.versions
+    system = platform.system()
+    print(f"Setting up {system} with Blender {', '.join(BLENDER_VERSIONS)}")
+    print(f"Using {'public URLs' if USE_PUBLIC_URLS else 'S3 bucket'}")
+    if system == "Linux":
+        setup_linux(install_x11=args.install_x11)
+    else:
+        {"Windows": setup_windows, "Darwin": setup_macos}[system]()
```
jusiskin: I wonder if this could be made more simple for maintainability:

```suggestion
    elif system == "Windows":
        setup_windows()
    elif system == "Darwin":
        setup_macos()
    else:
        raise OSError(f"Unsupported OS: {system}")
```

pipeline/setup-runner.py
```diff
@@ -0,0 +1,319 @@
...
+        "--versions", nargs="+", help="Blender versions to install (e.g., 4.5.4 4.2.12)"
+    )
+    parser.add_argument(
+        "--install-x11",
+        action="store_true",
+        help="Install X11 libraries and start Xvfb (Linux only)",
+    )
+    args = parser.parse_args()
+
+    USE_PUBLIC_URLS = args.public_urls
```
jusiskin, line 461: We should probably declare that this is a [`global` reference](https://docs.python.org/3/reference/simple_stmts.html#global):

```suggestion
    global USE_PUBLIC_URLS
    USE_PUBLIC_URLS = args.public_urls
```

pipeline/setup-runner.py
```diff
@@ -0,0 +1,319 @@
...
+        blender_exe = f"C:/Tools/blender-{version}-windows-x64/blender.exe"
+        run(
+            f'"{blender_exe}" --background --python installer\\add_submitter_to_pref.py -- --deadline_cloud_install_path {cwd}\\DeadlineCloudSubmitter\\Submitters\\Blender\\python',
+            shell=True,
+        )
+
+
+def setup_macos():
+    run("pip install --upgrade pip hatch", shell=True)
+    run("pip install --upgrade -r requirements-development.txt", shell=True)
```
jusiskin: Why are these here and not in the hatch environment's `pre-install` section?

pipeline/setup-runner.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Setup Blender test environment")
+    parser.add_argument(
+        "--public-urls", action="store_true", help="Download from public URLs instead of S3"
+    )
+    parser.add_argument(
+        "--versions", nargs="+", help="Blender versions to install (e.g., 4.5.4 4.2.12)"
+    )
```
jusiskin, line 450: I may be missing it, but it looks like these are being ignored

pipeline/setup-runner.py
```diff
@@ -0,0 +1,319 @@
...
+        run(f"rm -rf {submitter_path}", shell=True, check=False)
+
+    Path("./DeadlineCloudSubmitter/Submitters/Blender/python").mkdir(parents=True, exist_ok=True)
+    run(
+        "cp -r src/deadline/blender_submitter/addons/ ./DeadlineCloudSubmitter/Submitters/Blender/python/addons",
+        shell=True,
+    )
+
+    run(
+        'pip install --upgrade --python-version 3.11 --only-binary=:all: "deadline[gui]" blender-qt-stylesheet -t ./DeadlineCloudSubmitter/Submitters/Blender/python/modules',
```
jusiskin: Does the `--python-version` value here depend on the target version of Blender?

jusiskin (changes requested):

pipeline/setup-runner.py
moorec-aws: Yes

pipeline/setup-runner.py
moorec-aws: They were required for `hatch build`, I might not need them anymore tho. Will double check.

pipeline/setup-runner.py
moorec-aws: https://github.com/aws-deadline/deadline-cloud-for-blender/pull/296/files#diff-d345a893ea70e0d58995b9b5910fc51a9f5fc667b91a9502b538f1b05722267cR310

.github/workflows/integration_tests.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     permissions:
       id-token: write
       contents: read
-    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
+    uses: moorec-aws/.github/.github/workflows/reusable_integration_test.yml@mainline
```
moorec-aws: Will fix before merge

.github/workflows/integration_tests.yml
```diff
@@ -21,7 +21,7 @@ jobs:
...
```
moorec-aws: Will fix before merge

.github/workflows/integration_tests.yml
```diff
@@ -33,7 +33,7 @@ jobs:
...
```
moorec-aws: Will fix before merge

pipeline/setup-runner.py
moorec-aws: This is passed in now

pipeline/setup-runner.py
moorec-aws: The lines have been removed, not needed.

pipeline/setup-runner.py
moorec-aws: By default variables listed at the top of the file outside a function or class are global.

pipeline/setup-runner.py
moorec-aws: Simplified for maintainability.

hatch.toml
moorec-aws: Now one line.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/blender_utils.py
moorec-aws: good call, made this change.

test/integ/conftest.py
moorec-aws: Added some documentation and the validation.

pipeline/setup-runner.py
jusiskin: Sorry, I was incorrect here. I believe I had assumed this was inside a function, but it is only nested in:

```py
if __name__ == "__main__":
```

pipeline/setup-runner.py
```diff
@@ -0,0 +1,330 @@
...
+    return True
+
+
+def setup_linux(python_version, install_x11=False):
+    pkg_mgr = (
+        "dnf"
+        if subprocess.run("command -v dnf", shell=True, capture_output=True).returncode == 0
+        else "yum"
+    )
+    run(f"{pkg_mgr} update -y", shell=True)
```
jusiskin: As a best-practice, we should not be using `shell=True` and instead supply arguments as a list of `str`s. The reasons:

1. Avoids the need for an extra a shell process in between
2. Avoids needing to think about the proper use of escaping when arguments contains special shell characters (e.g. spaces, quotes, `*`)
3. It helps reduce the chance of a future command injection attack when we put a dynamic value into the command string

```suggestion
    run([pkg_mgr, "update", "-y"])
```

Same applies to the code below.

pipeline/setup-runner.py
```diff
@@ -0,0 +1,330 @@
...
+            continue
+
+        lock_file.touch()
+        try:
+            print(f"Installing Blender {version}...")
+            blender_archive = Path(f"/tmp/blender-{version}.tar.xz")
+            if not USE_PUBLIC_URLS and download_from_s3(
+                f"blender/blender-{version}-linux-x64.tar.xz", blender_archive
+            ):
+                pass
```
jusiskin: Should we also verify checksums here?

pipeline/setup-runner.py
```diff
@@ -0,0 +1,330 @@
...
+            continue
+
+        lock_file = Path(f"/tmp/blender-{version}.lock")
+        if lock_file.exists():
+            print(f"Waiting for concurrent Blender {version} install...")
+            for _ in range(60):
+                time.sleep(1)
+                if blender_marker.exists():
+                    break
+            continue
```
jusiskin, line 147: Can we document the lock files and the marker files with code comments? Maybe just to explain their goal?

jusiskin (changes requested):

pipeline/setup-runner.py
moorec-aws: Added comments

pipeline/setup-runner.py
moorec-aws: Removed all cases of `shell=true`

pipeline/setup-runner.py
moorec-aws: Probably should, I added validation for the version input as well.

pipeline/setup-runner.py
```diff
@@ -0,0 +1,491 @@
...
+                "libXinerama",
+                "libxkbcommon",
+                "libSM",
+                "libICE",
+                "libXt",
+                "libXmu",
+            ]
+        )
+        if run(["pgrep", "Xvfb"], check=False).returncode != 0:
+            run(["Xvfb", ":99", "-screen", "0", "1024x768x24"], check=False)
```
jusiskin: Before we had:

```py
run("Xvfb :99 -screen 0 1024x768x24 &")
```

note the trailing `&` here which detaches the process into the background. But now we do not have that behavior and I believe `run()` is synchronous and will wait for the `Xvfb` subprocess to exit.

test/integ/conftest.py
jusiskin: This code should probably be moved into the `blender_location` fixture rather than manipulating the `BLENDER_EXECUTABLE` env var

jusiskin (changes requested):

pipeline/setup-runner.py
moorec-aws: I switched to using subprocess.

test/integ/conftest.py
moorec-aws: Move to blender_location fixture.

justinsaws:
Just confirming that the commit history is going to get cleaned up before merge/your're squashing and cleaning up the message?

justinsaws (dismissed):

jusiskin (dismissed):

justinsaws (approved):

jusiskin (approved):

Issue #268, opened 2025-08-27
Title: Camera binding ("Bind Camera to Markers") is not respected when "All renderable cameras" option is used or when default camera is locked


fast-blender-add-ons-dot-com:
Title:
Camera binding ("Bind Camera to Markers") is not respected when "All renderable cameras" option is used or when default camera is locked

Body:

**Issue Summary**  
When submitting render jobs via the AWS Deadline Cloud Blender submitter, camera switches driven by "Bind Camera to Markers" are not respected. The render stays locked to the camera set by the marker at the initial frame, even when a marker later in the sequence should change the camera.

### Expected Behaviour

**Expected Behavior**  
Camera switches should occur per frame as driven by markers—blender-cli render (`-a`) behaves correctly in this regard.

---

### Current Behaviour

**Actual Behavior**  
The submitter appears to bake in the camera at the first frame of job submission and fails to re-evaluate markers per frame. Even selecting "All renderable cameras" does not work as expected—camera stays stuck.

### Reproduction Steps

**Steps to Reproduce**  
1. In Blender, set up timeline markers bound to different cameras.  
2. Prepare the scene so that the initial frames are using Camera A, and after a specific frame the marker binds to Camera B.  
3. Submit the job using Deadline Cloud Blender submitter, choosing either:  
   - **"All renderable cameras"**, or  
   - the **default option** (which appears to lock the scene’s current camera).  
4. The farm render only ever uses Camera A, even on frames after the marker.

### Environment

**Environment**  
- Submitter version: v0.5.5  
- Blender versions tested: 4.5 (also earlier versions)  
- Platform: Windows (but likely affects other OSes too)

fast-blender-add-ons-dot-com:
Did anybody get a chance to see this, it seems like a big company like Amazon would have this monitored constantly... just a little heads up because you know of course everybody wants to see you guys be successful

edwards-aws:
Thanks for reporting this issue! I was unable to reproduce this on Blender 4.2. I'll try again with Blender 4.5 and report back.

edwards-aws:
Here's my timeline that I used

<img width="328" height="307" alt="Image" src="https://github.com/user-attachments/assets/2b96a459-e2ab-4fa7-92ce-7bafd7913cb2" />

[...]

PR #308, opened 2026-01-08
feat!: Added option to render with the scene's default camera.

erico-aws:

The way the Submitter and Adaptor handle cameras bound to timeline markers is strange and unintuitive. 

We only allow users to either choose a specific camera or all cameras. There is no option to use the scene default or use marker bound cameras. However, when a scene with marker bound cameras is rendered, it always obeys the markers, even though the output filename includes the chosen camera. So a user could select Camera004 and render not a single frame from that camera if the bound cameras are different. This is especially weird when you select render all cameras. With render all cameras selected, a Task is created for each camera but they all render the same shots using the camera markers. 

First reported: https://github.com/aws-deadline/deadline-cloud-for-blender/issues/268

I've added an option to the submitter to use the default camera. When that option is selected, the camera parameter is not set in the task definition and not included in the run-data. We will rely on Blender to set the camera. This allows camera markers to be rendered as expected. Also, the output filename will not include a camera name if this option is selected.

Additionally, I've changed the behaviour of the regular camera option. When a camera is selected, it will force Blender to render all frames with that camera, regardless of camera markers or scene defaults. This means that when selecting the render all cameras option, each task will use the corresponding camera.

Note, I didn't want to remove the markers or disable them when a camera is selected because other things could be tied to them. The markers will override the scene's camera so, when setting a specific camera, we iterate over all markers and set the camera to be the chose camera. 

image

This change will now allow users to:

- Render with markers in an expected fashion
- Render with a specific (or all) camera(s) even with markers

Built the submitter and adaptor locally. 

Using the Submitter I exported a job bundle with the new camera option selected. The template did not contain a task parameter of the camera and camera was not defined in the run-data file. A unit test was added to ensure this.

I then ran a range of renders through the adaptor using init and run data files. 

Specifying a camera in run data always results in that camera rendering, no matter the markers or other scene defaults.
Without specifying a camera, the frames rendered used the scene default camera or cameras bound to markers.

```

test/integ/test_blender_adaptors.py::TestAdaptors::test_minimal_scene_adaptor 
[gw0] [ 25%] PASSED test/integ/test_blender_adaptors.py::TestAdaptors::test_minimal_scene_adaptor 
test/integ/test_blender_adaptors.py::TestAdaptors::test_minimal_scene_with_host_requirements_adaptor 
[gw0] [ 50%] PASSED test/integ/test_blender_adaptors.py::TestAdaptors::test_minimal_scene_with_host_requirements_adaptor 
test/integ/test_blender_submitters.py::TestSubmitters::test_minimal_scene_submitter 
[gw0] [ 75%] PASSED test/integ/test_blender_submitters.py::TestSubmitters::test_minimal_scene_submitter 
test/integ/test_blender_submitters.py::TestSubmitters::test_minimal_scene_with_host_requirements_submitter 
[gw0] [100%] PASSED test/integ/test_blender_submitters.py::TestSubmitters::test_minimal_scene_with_host_requirements_submitter 

================================================================================================================================================================================= slowest 5 durations ==================================================================================================================================================================================
45.63s call     test/integ/test_blender_adaptors.py::TestAdaptors::test_minimal_scene_adaptor
18.04s call     test/integ/test_blender_adaptors.py::TestAdaptors::test_minimal_scene_with_host_requirements_adaptor
3.52s call     test/integ/test_blender_submitters.py::TestSubmitters::test_minimal_scene_submitter
2.95s call     test/integ/test_blender_submitters.py::TestSubmitters::test_minimal_scene_with_host_requirements_submitter
0.61s setup    test/integ/test_blender_adaptors.py::TestAdaptors::test_minimal_scene_adaptor
============================================================================================================================================================================= 4 passed in 73.95s (0:01:13) =============================================================================================================================================================================
```

Yes. The schema for the run-data file has been loosened, which isn't a breaking change but the behaviour of the camera options is now very different. 
----

### DIFF
--- a/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/scene_settings_widget.py
+++ b/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/scene_settings_widget.py
@@ -32,6 +32,7 @@ _logger = logging.getLogger(__name__)
 # Default values for combo boxes in the addon UI.
 COMBO_DEFAULT_ALL_RENDERABLE_CAMERAS = "All Renderable Cameras"
 COMBO_DEFAULT_ALL_RENDERABLE_LAYERS = "All Renderable Layers"
+COMBO_USE_CAMERA_MARKERS = "Use Default Camera"
 
 
 class SceneSettingsWidget(QWidget):
@@ -161,6 +162,8 @@ class SceneSettingsWidget(QWidget):
             self.cameras_box.addItem(
                 COMBO_DEFAULT_ALL_RENDERABLE_CAMERAS, COMBO_DEFAULT_ALL_RENDERABLE_CAMERAS
             )
+            self.cameras_box.addItem(COMBO_USE_CAMERA_MARKERS, COMBO_USE_CAMERA_MARKERS)
+
             for cam in selectable_cameras:
                 self.cameras_box.addItem(cam, cam)
 
--- a/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/scene_settings_widget.py
+++ b/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/scene_settings_widget.py
@@ -32,7 +32,7 @@ _logger = logging.getLogger(__name__)
 # Default values for combo boxes in the addon UI.
 COMBO_DEFAULT_ALL_RENDERABLE_CAMERAS = "All Renderable Cameras"
 COMBO_DEFAULT_ALL_RENDERABLE_LAYERS = "All Renderable Layers"
-COMBO_USE_CAMERA_MARKERS = "Use Default Camera"
+COMBO_USE_DEFAULT_CAMERA = "Use Default Camera"
 
 
 class SceneSettingsWidget(QWidget):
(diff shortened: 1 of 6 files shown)


lucaseck (dismissed):

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/scene_settings_widget.py
```diff
@@ -32,6 +32,7 @@
 # Default values for combo boxes in the addon UI.
 COMBO_DEFAULT_ALL_RENDERABLE_CAMERAS = "All Renderable Cameras"
 COMBO_DEFAULT_ALL_RENDERABLE_LAYERS = "All Renderable Layers"
+COMBO_USE_CAMERA_MARKERS = "Use Default Camera"
```
epmog: Should the constant name be updated to reflect its value?

epmog (changes requested):
I realize we're addressing a potential blender usability issue by forcing the camera to be used even if there's markers that specify others. But we may want to revisit this in the future to just match the cli behaviour such that users aren't surprised by the behaviour.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/scene_settings_widget.py
erico-aws: You know, I had it as "Use Camera Markers" at first, ended up changing the text cause I felt it suited the functionality better. I've updated the constant name now.

epmog (approved):

lucaseck (approved):

PR #330, opened 2026-03-25
chore: fix CI test setup and add Blender 5.0 test support

AlexTranAmz:
- Blender integ test CI fail on Windows with code 4 after `xcopy` command
- We did not test Blender 5.0

- Code 4 fail due to xcopy did not handle path > MAX_PATH even if we enable it in the registry, so we need to shorten the path
   - After fixing it, we found out that temp folder also got changes so our folder become part of temp -> which is ignored by Blender hence the fix to include it as well 
- Add Blender 5.0 in test setup and test matrix

- Test start to run on Windows again
- We have Blender 5.0 test

- Deploy codebuild and run against this branch with all tests pass

Yes

No

----


seant-aws (dismissed):
nice !

hatch.toml
```diff
@@ -56,8 +56,13 @@ python = ["3.11"]
 blender = ["4.2.12"]
 python = ["3.11"]
 
+[[envs.integ-ci.matrix]]
+blender = ["5.0.1"]
+python = ["3.11"]
+
 [envs.integ-ci.env-vars]
 BLENDER_VERSION = "{matrix:blender}"
+BLENDER_TEMP_ALLOWLIST = "{root}"
```
godobyte, line 65: Q - what is this for, where is it used?

pipeline/setup-runner.py
```diff
@@ -6,16 +6,16 @@
...
 import os
 import platform
-import shutil
 import subprocess
 import sys
 import time
 from pathlib import Path
 
-BLENDER_VERSIONS = ["4.2.12", "4.5.4"]
+BLENDER_VERSIONS = ["4.2.12", "4.5.4", "5.0.1"]
```
godobyte, line 14: I would also indicate this change in the commit message/title.

pipeline/setup-runner.py
```diff
@@ -274,15 +277,7 @@ def setup_windows(python_version):
...
         major_minor = ".".join(version.split(".")[:2])
-        submitter_path = f"DeadlineCloudSubmitter\\Submitters\\Blender{major_minor}"
-
-        Path(f"{submitter_path}\\python").mkdir(parents=True, exist_ok=True)
-
-        # Remove stale addons directory from previous runs on reserved capacity instances.
-        # Windows xcopy fails with exit code 4 when copying into an existing destination.
-        addons_path = Path(f"{submitter_path}\\python\\addons")
-        if addons_path.exists():
-            shutil.rmtree(addons_path)
```
godobyte, line 285: What are the addons used for, would it change between runs? I know we didn't have it before, but wonder whether it actually has value considering the runner is reused.

hatch.toml
AlexTranAmz: Our current folder is marked as temp and Blender ignore it, so we have to add it to the allow list to be discovered by Blender

pipeline/setup-runner.py
AlexTranAmz: Done!

pipeline/setup-runner.py
AlexTranAmz: This addons is part of installing the submitter for Blender for testing, and will be changes. We did not need it anymore since as mention above, we're in a temp folder and will be clean up after run so no need to clean it

hatch.toml
AlexTranAmz: This is part of the submitter code [here](https://github.com/aws-deadline/deadline-cloud-for-blender/blob/d8f9504ef1cfaebbaf14c0a3cf14a8997c2fa076/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/blender_utils.py#L80)

seant-aws (approved):

godobyte (approved):

PR #331, opened 2026-03-25
feat: bundle GUI dependencies in the installer

crowecawcaw:
We want to install Qt in the installer instead of requiring a install step during the first run of the submitter.

Bundle it in the installer. Copy the allowlist of Qt files to bundle from deadline-cloud and apply it during the installer creation.

Submitter can be installed offline.

- I ran the installer and verified the installed plugin worked correctly and did not prompt to install GUI dependencies
- I inspected the install folder to make sure no unexpected Qt modules were included
- installer unit test

After updating the extension building script, I also:
- generated the extension, installed it, and verified it worked
- looked in the installed extension folder and verified it had the THIRD_PARTY_LICENSES files for all 3 platforms
- expanded a PySide wheel and verified it had only the allowlisted modules

n/a

```
    =================== 6 passed, 9 skipped, 1 warning in 44.35s ===================                                                                                                           
```

No update needed.

No

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,8 +32,13 @@ classifiers = [
 # Blender 4.1+ uses Python version 3.11
 
 dependencies = [
-    "deadline[gui] >= 0.54,< 0.55", 
+    "deadline >= 0.54,< 0.55",
     "openjd-adaptor-runtime >= 0.7,< 0.10",
+]
+
+[project.optional-dependencies]
+gui = [
+    "deadline[gui] >= 0.54,< 0.55",
 ]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
 # Blender 4.1+ uses Python version 3.11
 
 dependencies = [
-    "deadline >= 0.54,< 0.55", 
+    "deadline[gui] >= 0.54,< 0.55", 
     "openjd-adaptor-runtime >= 0.7,< 0.10",
 ]
 
--- a/scripts/build_addon.py
+++ b/scripts/build_addon.py
@@ -7,9 +7,12 @@ import os
 from pathlib import Path
 import shutil
 import subprocess
+import tarfile
 from tempfile import TemporaryDirectory
 import hashlib
 import re
+
+from depsBundle import _strip_pyside6
 
 parser = argparse.ArgumentParser(description="Experimental: Builds a Blender extension")
 parser.add_argument("--version", required=False)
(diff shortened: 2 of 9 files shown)


crowecawcaw:
Finished testing on Mac, Windows, and Linux, all working.

karthikbekalp (dismissed):
Used this PR for Cinema 4D: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/417 and it worked great! Thanks for the PR!

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/__init__.py
```diff
@@ -47,9 +45,6 @@ def execute(self, context):
 
         See the bpy Operator docs: https://docs.blender.org/api/current/bpy.types.Operator.html
         """
-        if not self.has_gui_deps():
-            self.install_gui()
```
epmog, line 51: Users can still `pip install deadline-cloud-for-blender` and set up blender, I wonder if we should make the gui install non-optional now... Or maybe just leave this in for those cases atm...

How does this change affect if you do the repository set-up pointing at the release page?

karthikbekalp: For Cinema 4D I added PySide [as a required dependency](https://github.com/karthikbekalp/deadline-cloud-for-cinema-4d/blob/f16eeef8d8e11648c627e2b78553d42fb45c1ba7/pyproject.toml#L37) as it does not make a lot of sense to use the submitter without GUI. 

Should we add the attributions to THIRD_PARTY_LICENSES now even though its a transitive dependency from `deadline-cloud`?

karthikbekalp: Synced about this offline. 
- We will be adding a separate top-level third party licenses 
- There was also thoughts abut removing the experimental approach in the future.

scripts/build_addon.py
```diff
@@ -19,13 +22,35 @@
 
 SUPPORTED_PLATFORMS = [
     "win_amd64",
-    "manylinux_2_17_x86_64",
-    "macosx_11_0_arm64",
-    "macosx_10_9_x86_64",
+    "manylinux_2_28_x86_64",
+    "macosx_12_0_arm64",
+    "macosx_12_0_x86_64",
```
crowecawcaw, line 27: Bump platform for pyside support.

karthikbekalp (dismissed):

pyproject.toml
```diff
@@ -32,7 +32,7 @@ classifiers = [
 # Blender 4.1+ uses Python version 3.11
 
 dependencies = [
-    "deadline >= 0.54,< 0.55", 
+    "deadline[gui] >= 0.54,< 0.55",
```
epmog: Maybe leave this out for now, since it'll get packaged with the adaptor and bloat its size. IF we need this for the addon installation method, we can move to an optional dependency, and specify that in the build script.

epmog (changes requested):

pyproject.toml
crowecawcaw: Made it optional

pyproject.toml
```diff
@@ -32,10 +32,15 @@ classifiers = [
...
 dependencies = [
-    "deadline >= 0.54,< 0.55", 
+    "deadline >= 0.54,< 0.55",
     "openjd-adaptor-runtime >= 0.7,< 0.10",
 ]
 
+[project.optional-dependencies]
+gui = [
+    "deadline[gui] >= 0.54,< 0.55",
+]
```
karthikbekalp, line 42: Hmm, do we not need PySide?

crowecawcaw: It should be part of `deadline[gui]` and pulled in.

karthikbekalp: > It should be part of deadline[gui] and pulled in.

Hmm, I remember a conversation that we had offline about adding both `deadline[gui]` and `PySide` for being future proof and only take the GUI components that we have license for. Anyway, not a blocker. Approving the PR.

karthikbekalp (approved):

epmog (approved):

PR #342, opened 2026-05-05
chore: add cross-platform adaptor test scripts

AlexTranAmz:

When iterating on the Blender adaptor or a job template, there is no quick way to test an adaptor build against a job bundle without running a full worker agent setup. The 3ds Max repository has two PowerShell scripts for this purpose (`test-3dsmax-adapter-run.ps1` and `test-3dsmax-openjd-run.ps1`), but they are Windows-only and there is no equivalent for Blender. This forces Blender adaptor developers to either spin up a worker agent for every test or hand-craft the init-data / run-data / openjd invocations manually.

Port the two 3ds Max testing scripts to the Blender repository with cross-platform support. Each flow is provided as both a PowerShell (`.ps1`) script and a Bash (`.sh`) script so users can run on any of Linux, macOS, or Windows.

 - scripts/test-blender-adapter-run.{sh,ps1}
    Invokes the adaptor directly via `blender-openjd run` with init-data and run-data derived from the job bundle. Renders only the first frame; useful as a quick smoke test of the init/run/stop flow.

- scripts/test-blender-openjd-run.{sh,ps1}
    Invokes `openjd run` against the job bundle template with the full frame list expanded into per-task parameters; exercises the daemon-mode start/run/stop sequence end-to-end through the real template.

Both scripts:
- Resolve the Blender executable from `--blender`, `BLENDER_EXECUTABLE`, the PATH, or platform-specific default install locations.
- Derive Blender's bundled Python from the Blender install (`<install>/<X.Y>/python/bin/...`) and install the adaptor wheel into it, mirroring the 3ds Max pattern of targeting the DCC's own Python.
- Parse `parameter_values.yaml` and `template.yaml` (PyYAML for bash, powershell-yaml for pwsh), merging template `parameterDefinitions` defaults with explicit values so sparse bundles still produce valid init data.
- Strip `deadline:` scoped parameters and any queue-environment parameters (e.g. `CondaPackages`, `CondaChannels`, `NamedCondaEnv*`) that are not declared in the template's `parameterDefinitions`, avoiding `openjd run` rejection for unknown job parameters.
- Honour declared task parameter types (INT Frame vs STRING Frame).
- Support `--path-mapping-file` for path mapping rules.

Blender adaptor developers can now smoke-test an adaptor build against any job bundle in under two minutes, on any platform, without a worker agent. No production code paths are affected.

Verified end-to-end on macOS against a real customer bundle (BMW27.blend, Cycles + Metal, 1280x720):

- `test-blender-adapter-run.sh` rendered frame 1 in 67s (exit 0).
- `test-blender-openjd-run.sh` rendered frames 1-2 in 93s through a full OpenJD session (`Chunks run: 2`, exit 0).

N/A — the existing integration suite is unaffected. The new scripts are developer tooling under `scripts/` and are not imported by the adaptor or submitter.

No

No. Pure addition of new developer-tooling scripts.

----

### DIFF
--- a/scripts/test-blender-adapter-run.ps1
+++ b/scripts/test-blender-adapter-run.ps1
@@ -1,0 +1,517 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Blender Adaptor Testing Script (cross-platform PowerShell)
+# Tests the deadline-cloud-for-blender adaptor directly without worker agent setup.
+#
+# Works on Linux, macOS, and Windows using PowerShell 7+ (pwsh).
+
+param(
+    [string]$JobBundleDir,
+    [string]$WheelPath = "",
+    [string]$BlenderExecutable = "",
+    [string]$BlenderPython = "",
+    [int]$Step = 0,
+    [string]$PathMappingFile = "",
+    [switch]$SkipInstall,
+    [switch]$ShowOutput,
+    [switch]$Help
+)
+
+# ------------------------------------------------------------------
+# Help
+# ------------------------------------------------------------------
+if ($Help) {
+    Write-Host "=== Blender Adaptor Testing Script Help ===" -ForegroundColor Green
+    Write-Host ""
+    Write-Host "DESCRIPTION:" -ForegroundColor Yellow
+    Write-Host "  Tests the deadline-cloud-for-blender adaptor directly without worker agent setup."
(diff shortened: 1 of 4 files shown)


seant-aws (approved):
This is great, looks like were able to do multiple frames on the openjd script which is good. When I was previously using these scripts for 3dsMax, I found that it was useful to have multiple cameras too, or other stuff that may be worth testing like render layers, etc. Probably a future addition when needed though

leongdl (approved):

PR #366, opened 2026-07-02
feat: implement BlenderSubmitter on the unified BaseSubmitter

leon-li-inspire:

Blender could only be driven through its native GUI submitter. The unified `BaseSubmitter` ABC (aws-deadline/deadline-cloud#1245) defines a shared, DCC-agnostic submission contract so hosts can be driven programmatically (e.g. the AYON/headless path) using one common interface. Blender needed an implementation of that contract, without regressing the existing GUI submit flow.

Implement `BlenderSubmitter(BaseSubmitter)` and route the GUI through it so both paths build identical job bundles.

- **New `deadline_cloud_blender_submitter/submitter.py`** — `BlenderSubmitter` implements `get_settings` / `get_job_template` / `get_parameter_values` / `get_asset_references` (+ inherited `get_submission_context`). Imports `BaseSubmitter` / `BaseSubmitterSettings` from `deadline.client.api`; the settings field is `job_name`.
- **`BlenderSubmitterSettings` lives in `template_filling.py`** (alongside the builders that consume it) and is re-exported from `submitter.py`. The builders read it directly, so the module dependency flows one way (`submitter` → `template_filling`, no import cycle). The GUI adapts its `BlenderSubmitterUISettings` → `BlenderSubmitterSettings` at the submission boundary via `_ui_settings_to_submitter_settings` in `open_deadline_cloud_dialog.py`. Blender stays **stateless** (Maya's stateful scene-context refactor is Maya-specific and not mirrored here).
- **`blender_utils` shared helpers** (used by both the GUI and API paths so they resolve identically):
  - `get_active_scene_name()` returns the real active scene (a valid `bpy.data.scenes` key) for the API path; `get_scene_name()` stays `.blend`-filename-derived so the dialog's default **job name** is unchanged.
  - `get_render_engine()` maps Blender engine ids (`BLENDER_EEVEE` / `BLENDER_EEVEE_NEXT` / `CYCLES` / `BLENDER_WORKBENCH`) to the template's `RenderEngine` allowed values (`eevee` / `cycles` / `workbench`), and raises a clear `DeadlineOperationError` for unsupported (third-party) engines rather than forwarding an out-of-range value that CreateJob rejects opaquely.
  - `resolve_output_path()` / `resolve_output_file_prefix()` resolve the output **directory** (with scene → prefs → blend-dir fallbacks) and filename prefix.
  - `resolve_gpu_settings()` reads `enable_gpu` / `gpu_device` from the scene, gating `gpu_device` to the `"NONE"` sentinel unless the scene actually renders on GPU (mirrors the widget).
  - `classify_auto_detected_paths()` auto-detects external dependencies (textures / linked libs / UDIM tiles / OCIO) and attaches them as job inputs.

Naming tracks aws-deadline/deadline-cloud#1245 + aws-deadline/deadline-cloud-for-maya#435: `SubmitterAPI` → `BaseSubmitter`, `SubmitterSettings` → `BaseSubmitterSettings` (re-exported from `deadline.client.api`); settings field `name` → `job_name`; module `submitter_api.py` → `submitter.py`; class `BlenderSubmitterAPI` → `BlenderSubmitter` (matches Maya's `MayaSubmitter`).

- Blender can now be driven through the unified `BaseSubmitter` contract (enables the AYON/headless submission path) **in addition to** the existing native GUI submitter.
- The GUI submit flow is preserved: it now delegates bundle construction to `BlenderSubmitter`, so the GUI and API paths produce identical templates and parameter values from one set of builders.
- One intentional behavior change in the GUI flow: `GPUDevice` now emits the `"NONE"` sentinel (uppercase) for CPU scenes instead of `"None"`, matching `default_blender_template.yaml` and the adaptor's case-sensitive `CyclesHandler` check (previously `"None"` could slip onto the GPU path on a GPU-equipped worker).
- **Depends on** aws-deadline/deadline-cloud#1245 — shipped in `deadline` **0.60.2**; `pyproject.toml` pins `deadline >= 0.60.2, < 0.61` (0.60.0/0.60.1 do not export the base classes).

- **Unit suite** against `deadline` 0.60.2: **60 passed, 0 skipped**; `ruff` + `black` + `mypy` clean. CI (full Python matrix + CodeQL) is green on the current head.
- **GUI publish → BlenderSubmitter → CreateJob → 10/10 SUCCEEDED** render (10 PNGs) on a live farm.

Submitter integration suite (`test/integ/test_blender_submitters.py`, `-m submitter`) — Blender 5.1.2, `deadline` 0.60.2, Rocky Linux 9, `QT_QPA_PLATFORM=offscreen`. Each test launches Blender headless, drives `BlenderSubmitter` via `_create_bundle_internal` to export a job bundle, and validates the generated `template.yaml` with `openjd check`:

```
$ pytest --no-cov test/integ -vvv --numprocesses=1 -m submitter
platform linux -- Python 3.11.13, pytest-9.1.1, pluggy-1.6.0
configfile: pyproject.toml
1 worker [3 items]

test/integ/test_blender_submitters.py::TestSubmitters::test_minimal_scene_submitter PASSED                    [ 33%]
test/integ/test_blender_submitters.py::TestSubmitters::test_minimal_scene_with_host_requirements_submitter PASSED [ 66%]
test/integ/test_blender_submitters.py::TestSubmitters::test_gpu_scene_submitter PASSED                         [100%]

============================== 3 passed in 6.70s ==============================
```

`installer/` is **not** modified. One file is **added** to `src/` (`.../deadline_cloud_blender_submitter/submitter.py`); no files removed. The addon is packaged as a directory tree (the installer copies `src/deadline/blender_submitter/addons/` wholesale), so adding a module within the existing package does not change the installer's file list or require installer-test updates.

No user-facing docs change. The new module, shared helpers, and the settings-shape/naming changes are documented via docstrings and this PR description; the naming migration tracks aws-deadline/deadline-cloud#1245 and aws-deadline/deadline-cloud-for-maya#435.

No. The native GUI submit flow is preserved (it now delegates to `BlenderSubmitter` but produces identical bundles), and `BlenderSubmitter` is an additive API. It does raise the minimum `deadline` dependency to `>= 0.60.2` (for the `BaseSubmitter` base classes).

----

### DIFF
--- a/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/blender_utils.py
+++ b/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/blender_utils.py
@@ -8,6 +8,7 @@ import os
 from pathlib import Path
 
 import bpy
+from deadline.client.exceptions import DeadlineOperationError
 
 _logger = logging.getLogger(__name__)
 
@@ -17,21 +18,20 @@ def get_renderable_view_layers(saved_scene_name) -> list[str]:
     during rendering.
 
     Args:
-        saved_scene_name: The name of the scene.
+        saved_scene_name: The name of the scene (a key in ``bpy.data.scenes``).
     """
-    scene_name = bpy.data.scenes[saved_scene_name].name
-    layers = [layer.name for layer in bpy.data.scenes[scene_name].view_layers if layer.use]
-    return layers
+    scene = bpy.data.scenes[saved_scene_name]
+    return [layer.name for layer in scene.view_layers if layer.use]
 
 
 def get_renderable_cameras(saved_scene_name) -> list[str]:
     """Returns a list of all camera objects in the scene that are marked as renderable.
 
     Args:
-        saved_scene_name: The name of the scene.
+        saved_scene_name: The name of the scene (a key in ``bpy.data.scenes``).
     """
-    scene_name = bpy.data.scenes[saved_scene_name].name
-    camera_names = [cam.name for cam in bpy.data.scenes[scene_name].objects if cam.type == "CAMERA"]
+    scene = bpy.data.scenes[saved_scene_name]
+    camera_names = [obj.name for obj in scene.objects if obj.type == "CAMERA"]
     return [cam for cam in camera_names if not bpy.data.objects[cam].hide_render]
 
 
@@ -53,9 +53,29 @@ def get_scene_resolution(saved_scene_name):
 
 
 def get_scene_name() -> str:
-    """Construct and return a name for the current scene based on the currently opened `.blend` file."""
+    """Construct and return a name for the current scene based on the currently opened `.blend` file.
+
+    The dialog submitter uses this as the default *job* name
+    (``open_deadline_cloud_dialog.py``), so it must stay derived from the
+    ``.blend`` file name. Callers that need the active scene's name as a
+    ``bpy.data.scenes`` key (e.g. the unified ``BlenderSubmitter`` path) must
+    use ``get_active_scene_name()`` instead.
+    """
     scene_name = bpy.path.basename(bpy.context.blend_data.filepath).replace(".blend", "")
     return scene_name
+
+
+def get_active_scene_name() -> str:
+    """Return the name of the active Blender scene.
+
+    This is an actual key in ``bpy.data.scenes`` (unlike ``get_scene_name()``,
+    which derives a label from the ``.blend`` file name). Callers that index
+    ``bpy.data.scenes`` — e.g. ``get_renderable_view_layers`` /
+    ``get_renderable_cameras`` — must use this, since the active scene is not
+    necessarily named identically to the file (the default scene is ``"Scene"``),
+    which would otherwise raise a KeyError during job-template construction.
+    """
+    return bpy.context.scene.name
 
 
 def get_frames() -> str:
@@ -63,6 +83,117 @@ def get_frames() -> str:
     start = bpy.context.scene.frame_start
     end = bpy.context.scene.frame_end
     return str(start) + "-" + str(end)
+
+
+# Maps Blender's internal engine id (e.g. "BLENDER_EEVEE", "CYCLES") to the
+# render-engine names the job template's RenderEngine parameter allows
+# ("eevee"/"cycles"/"workbench"). Emitting the raw "blender_eevee" id fails
+# CreateJob with a ValidationException.
+_RENDER_ENGINE_MAP = {
+    "BLENDER_EEVEE": "eevee",
+    "BLENDER_EEVEE_NEXT": "eevee",
+    "CYCLES": "cycles",
+    "BLENDER_WORKBENCH": "workbench",
+}
+
+
+def get_render_engine() -> str:
+    """Return the active scene's render engine as a template RenderEngine value.
+
+    Shared by the GUI submitter and the unified API path so both map Blender's
+    internal engine id to the template's allowed values identically.
+
+    Raises:
+        DeadlineOperationError: If the active engine is not one of the built-in
+            engines the job template supports (Cycles / EEVEE / Workbench).
+            Third-party engines (V-Ray, Octane, Redshift, …) would otherwise be
+            forwarded as an out-of-range ``RenderEngine`` value that CreateJob
+            rejects with an opaque ValidationException, so fail early with an
+            actionable message instead.
+    """
+    engine = bpy.context.scene.render.engine
+    if engine not in _RENDER_ENGINE_MAP:
+        raise DeadlineOperationError(
+            f"Render engine '{engine}' is not supported. Switch the scene's render "
+            "engine to Cycles, EEVEE, or Workbench before submitting."
+        )
+    return _RENDER_ENGINE_MAP[engine]
+
+
+def resolve_output_path() -> str:
+    """Return the render output *directory* for the active scene.
+
+    The OutputDir job parameter is type PATH / objectType DIRECTORY, so this
+    must be a directory (``scene.render.filepath`` is a directory + filename
+    prefix, e.g. ``//render/frame_``). Falls back, in order, to:
+
+    1. the directory of ``scene.render.filepath`` (if it has a directory part),
+    2. the Preferences "render output directory", then
+    3. the directory containing the ``.blend`` file.
+
+    Shared by the GUI submitter and the unified API path so both resolve the
+    output directory (and its fallbacks) identically.
+    """
+    render_filepath = bpy.context.scene.render.filepath
+    if os.path.dirname(render_filepath):
+        return os.path.dirname(bpy.path.abspath(render_filepath))
+    if bpy.context.preferences.filepaths.render_output_directory:
+        return bpy.context.preferences.filepaths.render_output_directory
+    return os.path.dirname(bpy.context.blend_data.filepath)
+
+
+def resolve_output_file_prefix() -> str:
+    """Return the output filename prefix from ``scene.render.filepath``.
+
+    Empty when the scene's render filepath has no filename part (the caller
+    keeps its own default in that case). Shared by both submission flows.
+    """
+    return bpy.path.basename(bpy.context.scene.render.filepath)
+
+
+def resolve_gpu_settings() -> tuple[bool, str]:
+    """Return ``(enable_gpu, gpu_device)`` read from the scene/preferences.
+
+    ``gpu_device`` defaults to the ``"NONE"`` sentinel (uppercase) — matching
+    ``default_blender_template.yaml`` and the adaptor's ``CyclesHandler`` check
+    — and is set to the Cycles ``compute_device_type`` only when the scene is
+    actually rendering on GPU (``scene.cycles.device == "GPU"``).
+
+    The ``compute_device_type`` preference persists independently of the scene's
+    render device, so gating on ``enable_gpu`` here keeps ``gpu_device`` at
+    ``"NONE"`` for a CPU scene even when a GPU device type is selected in
+    Preferences — mirroring ``scene_settings_widget.update_settings`` so the GUI
+    and unified API paths enable GPU rendering identically.
+    """
+    enable_gpu = bpy.context.scene.cycles.device == "GPU"
+    gpu_device = "NONE"
+    if enable_gpu:
+        compute_device_type = bpy.context.preferences.addons[
+            "cycles"
+        ].preferences.compute_device_type
+        if compute_device_type != "NONE":
+            gpu_device = compute_device_type
+    return enable_gpu, gpu_device
+
+
+def classify_auto_detected_paths(project_path) -> tuple[set[str], set[str]]:
+    """Auto-detect external dependencies and split them into files/directories.
+
+    Runs :func:`find_files` for ``project_path`` and classifies each result as
+    an input filename or input directory. Shared by the GUI submitter and the
+    unified API path so both attach the same auto-detected assets.
+
+    Returns:
+        A ``(input_filenames, input_directories)`` tuple of string sets.
+    """
+    input_filenames: set[str] = set()
+    input_directories: set[str] = set()
+    for f in find_files(project_path):
+        if f.is_dir():
+            input_directories.add(str(f))
+        else:
+            input_filenames.add(str(f))
+    return input_filenames, input_directories
 
 
 def find_files(project_path, skip_temp=True, skip_nonexistent=True) -> list[Path]:
--- a/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/ocio_utils.py
+++ b/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/ocio_utils.py
@@ -14,19 +14,43 @@ def get_ocio_config(filename: str) -> ocio.Config:
     return ocio.Config.CreateFromFile(filename)
 
 
-def get_ocio_referenced_dirs(config: ocio.Config) -> list[str]:
+def get_ocio_referenced_dirs(config: ocio.Config, config_path: str = "") -> list[str]:
     """
     Given an OCIO Config object, parse its `search_path` attribute and
     return a list of referenced directories.
+
+    Relative search paths are resolved against ``config_path``'s directory (the
+    directory of the config file the search paths belong to), falling back to
+    the active ``$OCIO`` config directory when ``config_path`` is not given.
     """
+    base_dir = os.path.dirname(config_path or get_ocio_path())
 
     path_list = []
     if config.getSearchPath():
         for path in config.getSearchPaths():
             # Paths may be absolute or relative to the config file.
             if not Path(path).is_absolute():
-                path = Path.joinpath(Path(os.path.dirname(get_ocio_path())), path)
+                path = Path.joinpath(Path(base_dir), path)
 
             path_list.append(path)
 
     return path_list
+
+
+def get_current_ocio_referenced_dirs(ocio_path: str = "") -> list[str]:
+    """Return the directories referenced by an OCIO config, if any.
+
+    Reads the OCIO config at ``ocio_path`` (defaulting to the active config via
+    the ``OCIO`` env var when not given) and returns the directories it
+    references, so they can be added as job input attachments. Empty when no
+    OCIO config is set. Shared by the GUI submitter and the unified API path so
+    both attach the same OCIO-referenced directories.
+    """
+    ocio_path = ocio_path or get_ocio_path()
+    if not ocio_path:
+        return []
+    config = get_ocio_config(ocio_path)
+    # Resolve relative search paths against this config's own directory, not
+    # necessarily the $OCIO one (they can differ when a caller passes an
+    # explicit ocio_path).
+    return [str(path) for path in get_ocio_referenced_dirs(config, ocio_path)]
--- a/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
+++ b/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
@@ -1,8 +1,7 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
-import os
 from pathlib import Path
-from typing import Any, Optional
+from typing import Any, Optional, cast
 
 import bpy
 from qtpy.QtCore import Qt  # type: ignore
@@ -29,6 +28,7 @@ from . import sanity_checks as sc
 from . import scene_settings_widget as ssw
 from . import template_filling as tf
 from . import ocio_utils as ocio
+from .submitter import BlenderSubmitter, BlenderSubmitterSettings
 from ._version import version
 from ._version import version_tuple as adaptor_version_tuple
 
@@ -75,26 +75,14 @@ def create_deadline_dialog(parent=None) -> SubmitJobToDeadlineDialog:
     settings.project_path = bpy.context.blend_data.filepath
     settings.frame_list = bu.get_frames()
 
-    # For the output path, first check for a value in the Scene settings
-    if os.path.dirname(bpy.context.scene.render.filepath):
-        settings.output_path = os.path.dirname(bpy.path.abspath(bpy.context.scene.render.filepath))
-    # If none, use the one in Preferences
-    elif bpy.context.preferences.filepaths.render_output_directory:
-        settings.output_path = bpy.context.preferences.filepaths.render_output_directory
-    # If neither of these are set, use the job bundle directory by default
-    else:
-        settings.output_path = os.path.dirname(bpy.context.blend_data.filepath)
+    # Resolve the output directory + filename prefix via the shared helpers so
+    # the GUI and the unified API path stay in lock-step.
+    settings.output_path = bu.resolve_output_path()
+    if bu.resolve_output_file_prefix():
+        settings.output_file_prefix = bu.resolve_output_file_prefix()
 
-    if bpy.path.basename(bpy.context.scene.render.filepath):
-        settings.output_file_prefix = bpy.path.basename(bpy.context.scene.render.filepath)
-
-    # Read the user's preferences to set GPU settings.
-    settings.enable_gpu = bpy.context.scene.cycles.device == "GPU"
-
-    if bpy.context.preferences.addons["cycles"].preferences.compute_device_type != "NONE":
-        settings.gpu_device = bpy.context.preferences.addons[
-            "cycles"
-        ].preferences.compute_device_type
+    # Read GPU settings from the scene/preferences (shared helper).
+    settings.enable_gpu, settings.gpu_device = bu.resolve_gpu_settings()
 
     # Load and set sticky settings, if any.
     settings.load_sticky_settings(settings.project_path)
@@ -105,11 +93,10 @@ def create_deadline_dialog(parent=None) -> SubmitJobToDeadlineDialog:
     # Set auto-detected attachments.
     auto_detected_attachments = _get_auto_detected_assets(settings.project_path)
 
-    # If the user has an OCIO config file set, we will have to upload any directories referenced by that file as job attachments.
-    if ocio.get_ocio_path():
-        ocio_config = ocio.get_ocio_config(ocio.get_ocio_path())
-        for path in ocio.get_ocio_referenced_dirs(ocio_config):
-            auto_detected_attachments.input_directories.add(str(path))
+    # Add the directories referenced by a custom OCIO config (shared helper) as
+    # job attachments, so custom color management resolves on the farm.
+    for path in ocio.get_current_ocio_referenced_dirs():
+        auto_detected_attachments.input_directories.add(path)
 
     # Set regular attachments.
     attachments = AssetReferences(
@@ -210,6 +197,47 @@ def _create_bundle(
     )
 
 
+def _ui_settings_to_submitter_settings(
+    settings: tf.BlenderSubmitterUISettings,
+) -> BlenderSubmitterSettings:
+    """Adapt the GUI's BlenderSubmitterUISettings to the headless BlenderSubmitterSettings.
+
+    The GUI binds its Qt widgets to BlenderSubmitterUISettings, while the
+    builders now consume BlenderSubmitterSettings (the unified engine's shape),
+    so we convert at the submission boundary — the mirror of the old
+    ``_to_native_settings``. This is the single adaptation point; both the GUI
+    and the unified API path then run the same BlenderSubmitter builders.
+
+    The effective frame range is resolved here (override -> the user's range,
+    otherwise the scene range) so the engine sees a ready-to-use ``frame_list``.
+    """
+    frame_list = settings.frame_list if settings.override_frame_range else bu.get_frames()
+    width, height = bpy.context.scene.render.resolution_x, bpy.context.scene.render.resolution_y
+
+    return BlenderSubmitterSettings(
+        job_name=settings.name,
+        description=settings.description,
+        frame_list=frame_list,
+        project_path=settings.project_path,
+        output_path=settings.output_path,
+        override_frame_range=settings.override_frame_range,
+        input_filenames=list(settings.input_filenames),
+        input_directories=list(settings.input_directories),
+        output_directories=list(settings.output_directories),
+        renderer_name=settings.renderer_name,
+        scene_name=settings.scene_name,
+        view_layer_selection=settings.view_layer_selection,
+        camera_selection=settings.camera_selection,
+        image_width=width,
+        image_height=height,
+        output_file_prefix=settings.output_file_prefix,
+        enable_gpu=settings.enable_gpu,
+        gpu_device=settings.gpu_device,
+        ocio_config_path=settings.ocio_config_path,
+        include_adaptor_wheels=settings.include_adaptor_wheels,
+    )
+
+
 def _create_bundle_internal(
     widget: SubmitJobToDeadlineDialog,
     job_bundle_dir: str,
@@ -222,6 +250,11 @@ def _create_bundle_internal(
 ) -> dict[str, Any]:
     """Create and write Deadline job bundle files to the given directory. Also save sticky settings.
     The internal variation exists so that the functionality can be accessed without a UI.
+
+    The bundle itself is built by the unified :class:`BlenderSubmitter` (the
+    single source of truth shared with the AYON/headless path); this callback
+    only owns the GUI-specific concerns — sanity checks, adapting the UI
+    settings, writing the files, and saving sticky settings.
 
     Args:
         widget: The submitter dialog widget.
@@ -238,41 +271,16 @@ def _create_bundle_internal(
     # Run sanity checks on submission
     sc.run_sanity_checks(settings, prompt_for_saving)
 
-    renderable_cameras = bu.get_renderable_cameras(settings.scene_name)
-
-    common_layer_settings = tf.CommonLayerSettings(
-        renderer_name=settings.renderer_name,
-        frame_range=bu.get_frames(),
-        renderable_camera_names=renderable_cameras,
-        output_directories=settings.output_path,
-        output_file_prefix=settings.output_file_prefix,
-        image_resolution=(
-            bpy.context.scene.render.resolution_x,
-            bpy.context.scene.render.resolution_y,
-        ),
-        ui_group_label=settings.ui_group_label,
-        frames_parameter_name=settings.frames_parameter_name,
-        output_file_prefix_parameter_name=settings.output_file_prefix_parameter_name,
-        image_width_parameter_name=settings.image_width_parameter_name,
-        image_height_parameter_name=settings.image_height_parameter_name,
-        scene_name=settings.scene_name,
+    # Delegate bundle construction to the unified submitter so the GUI and the
+    # headless/AYON path build identical templates and parameter values.
+    submitter = BlenderSubmitter()
+    submitter_settings = _ui_settings_to_submitter_settings(settings)
+    job_template = submitter.get_job_template(submitter_settings, host_requirements)
+    # queue_parameters is list[JobParameter] (a TypedDict); the unified builder
+    # types it as list[dict] — same shape at runtime, so cast to bridge.
+    parameter_values = submitter.get_parameter_values(
+        submitter_settings, cast("list[dict[str, Any]]", queue_parameters)
     )
-
-    # Add selected layers to the list of layers to render.
-    layer_names: list[str] = []
-    if settings.view_layer_selection == ssw.COMBO_DEFAULT_ALL_RENDERABLE_LAYERS:
-        for layer in bu.get_renderable_view_layers(settings.scene_name):
-            layer_names.append(layer)
-    else:
-        layer_names.append(settings.view_layer_selection)
-
-    if settings.override_frame_range:
-        common_layer_settings.frame_range = settings.frame_list
-
-    job_template = tf.fill_job_template(
-        settings, layer_names, common_layer_settings, host_requirements
-    )
-    parameter_values = tf.get_parameter_values(settings, common_layer_settings, queue_parameters)
 
     # Write the job bundle files.
     job_bundle_path = Path(job_bundle_dir)
@@ -298,17 +306,9 @@ def _create_bundle_internal(
 
 
 def _get_auto_detected_assets(project_path: str) -> AssetReferences:
-    files = bu.find_files(project_path)
-
-    # Sort auto-detected attachments to classify files and directories correctly.
-    input_filenames = set()
-    input_directories = set()
-    for f in files:
-        if f.is_dir():
-            input_directories.add(str(f))
-        else:
-            input_filenames.add(str(f))
-
+    # Classify auto-detected dependencies via the shared helper so the GUI and
+    # the unified API path attach identical assets.
+    input_filenames, input_directories = bu.classify_auto_detected_paths(project_path)
     return AssetReferences(input_filenames=input_filenames, input_directories=input_directories)
 
 
--- a/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
+++ b/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
@@ -1,0 +1,244 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+from __future__ import annotations
+
+import logging
+from typing import Any, Optional, cast
+
+from deadline.client.api import BaseSubmitter, BaseSubmitterSettings
+from deadline.client.exceptions import DeadlineOperationError
+
+import bpy  # type: ignore[import]
+
+from deadline.client.job_bundle.parameters import JobParameter
+from deadline.client.job_bundle.submission import AssetReferences
+
+from . import ocio_utils as ocio
+from .blender_utils import (
+    classify_auto_detected_paths,
+    get_active_scene_name,
+    get_frames,
+    get_render_engine,
+    get_renderable_cameras,
+    get_renderable_view_layers,
+    resolve_gpu_settings,
+    resolve_output_file_prefix,
+    resolve_output_path,
+)
+from .template_filling import (
+    BlenderSubmitterSettings,
+    CommonLayerSettings,
+    fill_job_template,
+    get_parameter_values as _fill_parameter_values,
+)
+
+_logger = logging.getLogger(__name__)
+
+# Re-exported for callers that import it from this module (e.g. the GUI dialog
+# and tests). The definition lives in ``template_filling`` so the module
+# dependency flows one way and avoids an import cycle.
+__all__ = ["BlenderSubmitter", "BlenderSubmitterSettings"]
+
+# Sentinel for "render every renderable view layer". Duplicated from
+# scene_settings_widget.COMBO_DEFAULT_ALL_RENDERABLE_LAYERS to avoid importing
+# the Qt-heavy widget module into this headless engine (the same pattern
+# template_filling uses for the camera sentinels).
+_ALL_RENDERABLE_LAYERS = "All Renderable Layers"
+
+
+class BlenderSubmitter(BaseSubmitter):
+    """BaseSubmitter implementation for Blender submissions."""
+
+    def get_settings(self) -> BlenderSubmitterSettings:
+        settings = BlenderSubmitterSettings()
+        scene = bpy.context.scene
+
+        settings.job_name = bpy.path.basename(bpy.data.filepath) or "Untitled"
+        settings.project_path = bpy.data.filepath or ""
+        # scene_name must be a key in bpy.data.scenes (indexed downstream by
+        # get_renderable_view_layers/cameras), so use the active scene name
+        # rather than the .blend-derived label.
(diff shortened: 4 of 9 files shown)


src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,163 @@
...
+            scene_name=scene_name,
+        )
+
+    def get_asset_references(self, settings: SubmitterSettings) -> dict[str, Any]:
+        asset_refs = AssetReferences(
+            input_filenames={bpy.data.filepath} if bpy.data.filepath else set(),
+            input_directories=set(settings.input_directories),
+            output_directories=set(settings.output_directories),
+        )
+        return asset_refs.to_dict()
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in e9e1c39 — `get_asset_references` now runs `find_files(bpy.data.filepath)` auto-detection so external dependencies (textures / linked libs / OCIO) are attached as job inputs, best-effort (a scan failure doesn't block submission). Covered by `test_get_asset_references_includes_auto_detected` + `…_survives_scan_failure`.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter_api.py
```diff
@@ -0,0 +1,171 @@
...
+            "BLENDER_EEVEE": "eevee",
+            "BLENDER_EEVEE_NEXT": "eevee",
+            "CYCLES": "cycles",
+            "BLENDER_WORKBENCH": "workbench",
+        }
+        settings.renderer_name = _engine_map.get(scene.render.engine, scene.render.engine.lower())
+
+        settings.image_width = scene.render.resolution_x
+        settings.image_height = scene.render.resolution_y
+        settings.output_path = (
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in e9e1c39 — `output_path` is now `os.path.dirname(bpy.path.abspath(scene.render.filepath))` (the render directory), with the filename carried in `output_file_prefix`, matching the dialog flow.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,208 @@
...
+from deadline.client.submitter_api import SubmitterAPI, SubmitterSettings
+
+import bpy  # type: ignore[import]
+
+from deadline.client.job_bundle.submission import AssetReferences
+
+if TYPE_CHECKING:
+    # Only referenced in a string cast() annotation below; import under
+    # TYPE_CHECKING so it is not an unused runtime import (CodeQL).
+    from deadline.client.job_bundle.parameters import JobParameter
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in e9e1c39 — the `JobParameter` import is only used in a string annotation, so it moved under `TYPE_CHECKING`, clearing the CodeQL finding.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,206 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+from __future__ import annotations
```
leongdl, line 2: Question - this seems to be code duplication? Will we integrate this class with the normal submitter code paths?

leongdl: I'm thinking, can this class become a model class used by the submitter, then all UI code stays in the UI, all states and selections that affect a job goto this class, and the API of this class it to build a job bundle.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter_api.py
```diff
@@ -0,0 +1,208 @@
...
+            renderer_name = settings.renderer_name
+            scene_name = settings.scene_name
+            output_file_prefix = settings.output_file_prefix
+            image_resolution = (settings.image_width, settings.image_height)
+
+        return CommonLayerSettings(
+            renderer_name=renderer_name,
+            frame_range=settings.frame_list,
+            frames_parameter_name="Frames",
+            renderable_camera_names=[],
```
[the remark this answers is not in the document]
leongdl: This seems to be a bug?

leongdl:
Following up on [my comment above](https://github.com/aws-deadline/deadline-cloud-for-blender/pull/366#discussion_r3581876998) about making this class a model used by the submitter — expanding with a concrete proposal. The clean layering would be **GUI / API → Data model → API class → Output**:

```
Sources (write to model):   GUI dialog + scene_settings_widget, or external API callers
        │
Data model:                 BlenderJobModel (evolved from BlenderSubmitterUISettings)
                            - pure data: engine, frames, output, layers, cameras, GPU, OCIO, attachments
                            - from_scene() factory = the ONLY place that reads bpy state
                            - sticky settings load/save
        │
API class (behavior):       BlenderSubmitterAPI(SubmitterAPI) wraps a model instance
                            - get_job_template / get_parameter_values / get_asset_references
                            - convenience build_bundle()
        │
Output:                     job bundle (template.yaml, parameter_values.yaml, asset_references.yaml)
```

Design rules that keep it clean:

1. **The model is dumb data.** No `bpy` calls on instance methods — only `from_scene()` touches Blender. That keeps the model testable and constructible by callers not running inside Blender.
2. **Build on `BlenderSubmitterUISettings`, don't add a parallel class.** It already holds every job-affecting field plus sticky persistence — it *is* 80% of this model. `BlenderSubmitterSettings` in this PR duplicates its defaults and bridges via `_to_native_settings`; with one model, the widget mutations and API setters are literally the same fields, so the translation layer and its drift risk disappear. `CommonLayerSettings` becomes an internal derivation.
3. **Submit-time inputs are arguments, not fields.** `queue_parameters` and `host_requirements` come from the dialog framework (or API caller) at submission time — pass them into the build methods rather than storing them on the model.
4. **Keep the three build methods separate** (matching the `SubmitterAPI` ABC shape) plus a composed `build_bundle()`, so this class collapses to a thin adapter.

Why it's worth it: today the scene-gathering logic is duplicated between `open_deadline_cloud_dialog.py` and this file, and the copies have already diverged (output-path fallbacks, OCIO attachment dirs, GPU detection, renderable cameras). A single `from_scene()` kills that whole class of bug.

Suggested sequencing so this PR isn't blocked on a big refactor:

1. **This PR:** extract the shared gathering helpers (output path/prefix resolution, asset auto-detection incl. OCIO, engine map) into `blender_utils` and call them from both flows — fixes the behavioral drift now.
2. **Follow-up:** absorb the settings classes into the model with `from_scene()`, re-point the dialog callback at the API class, delete `_to_native_settings`.

Everything extracted in step 1 becomes the body of `from_scene()` later, so no work is thrown away. If this pattern works here it's also the template for the other DCC submitters.

test/unit/deadline_submitter_for_blender/test_submitter.py
```diff
@@ -513,3 +513,111 @@ def test_fill_job_template_use_default_camera(submitter_settings, common_layer_s
     assert (
         "camera:" not in run_data
     ), "Camera should not be set in RunData when using default camera"
+
+
```
leon-li-inspire, line 127: Rename this test file from "test_submitter.py" to "test_gui_submitter.py"

test/unit/deadline_submitter_for_blender/test_submitter_engine.py
```diff
@@ -0,0 +1,332 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""Tests for the unified BlenderSubmitter (…deadline_cloud_blender_submitter.submitter).
```
leon-li-inspire: Rename this file from "test_submitter_engine.py" to "test_submitter.py"

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,263 @@
...
+            pass
+
+        asset_refs = AssetReferences(
+            input_filenames=input_filenames,
+            input_directories=input_directories,
+            output_directories=set(settings.output_directories),
+        )
+        return asset_refs.to_dict()
+
+    def _to_native_settings(self, settings: BaseSubmitterSettings):
```
leon-li-inspire: Why do we need this method? Shouldn't just need `_ui_settings_to_submitter_settings` ? And in "blender_settings = self._to_native_settings(settings)" of "get_parameter_values", could we just cast the settings to `BlenderSubmitterSetting`?

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/blender_utils.py
```diff
@@ -54,9 +54,15 @@ def get_scene_resolution(saved_scene_name):
...
-    return scene_name
+    """Return the name of the active Blender scene.
+
+    This must be an actual scene name in ``bpy.data.scenes`` because callers
+    (e.g. ``get_renderable_view_layers``/``get_renderable_cameras``) index
+    ``bpy.data.scenes`` by it. Deriving it from the ``.blend`` file name breaks
+    whenever the scene is not named identically to the file (the default scene
+    is ``"Scene"``), raising a KeyError during job-template construction.
+    """
+    return bpy.context.scene.name
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in 249a37d (and the delegation refactor before it). `get_scene_name()` is back to the `.blend`-filename form the dialog uses for the default job name; the API path now uses a separate `get_active_scene_name()` (returns `bpy.context.scene.name`) for the `bpy.data.scenes` key. Covered by `TestSceneName` in `test_blender_utils.py`.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter_api.py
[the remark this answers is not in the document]
leon-li-inspire: Fixed. `_build_common_layer_settings` now populates `renderable_camera_names` via `get_renderable_cameras(scene_name)` (best-effort — a bad scene name won't block template building), so the "All Renderable Cameras" selection produces a real Camera dimension. Covered by `test_build_common_layer_settings_populates_renderable_cameras`.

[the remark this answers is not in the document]
leon-li-inspire: Yes — that hardcoded `renderable_camera_names=[]` was a bug. Fixed: it's now populated from `get_renderable_cameras(scene_name)` (best-effort). Covered by `test_build_common_layer_settings_populates_renderable_cameras` + `…_survives_bad_scene_name`.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter_api.py
```diff
@@ -0,0 +1,208 @@
...
+class BlenderSubmitterAPI(SubmitterAPI):
+    """SubmitterAPI implementation for Blender submissions."""
+
+    def get_settings(self) -> BlenderSubmitterSettings:
+        settings = BlenderSubmitterSettings()
+        scene = bpy.context.scene
+
+        settings.name = bpy.path.basename(bpy.data.filepath) or "Untitled"
+        settings.project_path = bpy.data.filepath or ""
+        settings.scene_name = get_scene_name()
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed. `get_settings()` now sets `ocio_config_path` from `ocio.get_ocio_path()`, and `get_asset_references` attaches the OCIO-referenced dirs via the shared `ocio.get_current_ocio_referenced_dirs(...)` helper (honoring an explicitly-set path, else the active config; best-effort). Covered by `test_get_asset_references_attaches_ocio_dirs` + `…_survives_bad_ocio_config` and `test_get_settings_populates_ocio_config_path`.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,206 @@
...
+        else:
+            settings.output_path = ""
+
+        settings.input_filenames = [bpy.data.filepath] if bpy.data.filepath else []
+        settings.output_directories = [settings.output_path] if settings.output_path else []
+
+        if scene.view_layers:
+            settings.view_layer_selection = scene.view_layers[0].name
+
+        if scene.camera:
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed. When the scene has no camera, `get_settings()` now defaults `camera_selection` to `"Use Default Camera"` instead of leaving `""` (which produced the bogus `[""]` camera dimension). Covered by `test_get_settings_defaults_camera_when_none`.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,206 @@
...
+            ui_group_label="Blender Settings",
+            image_width_parameter_name="ResolutionX",
+            image_height_parameter_name="ResolutionY",
+            image_resolution=image_resolution,
+            scene_name=scene_name,
+        )
+
+    def get_asset_references(self, settings: SubmitterSettings) -> dict[str, Any]:
+        input_filenames: set[str] = set()
+        input_directories: set[str] = set(settings.input_directories)
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed. `get_asset_references` now seeds `input_filenames` from `settings.input_filenames` (the same way it seeds `input_directories`), so caller-supplied input files are honored. Covered by `test_get_asset_references_honors_caller_input_filenames`.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,243 @@
...
+    def get_job_template(
+        self,
+        settings: BaseSubmitterSettings,
+        host_requirements: Optional[dict[str, Any]] = None,
+    ) -> dict[str, Any]:
+        blender_settings = self._to_native_settings(settings)
+        # Use the actual scene name (settings.job_name is the .blend file name,
+        # which is not a key in bpy.data.scenes and raises KeyError here).
+        scene_name = getattr(settings, "scene_name", None) or settings.job_name
+        view_layers = get_renderable_view_layers(scene_name)
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in the delegation refactor (18526a8). `get_job_template` now calls `_resolve_view_layer_names(settings)`, which honors `view_layer_selection`: empty or the "All Renderable Layers" sentinel → every renderable layer; otherwise just the selected layer — matching the dialog. `get_settings()` now defaults the selection to "" (all renderable) rather than pinning `view_layers[0]`. Covered by `test_resolve_view_layers_*` in `test_submitter.py`.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
leon-li-inspire: Addressed. The GUI dialog no longer builds the bundle on a parallel path — it now delegates to the unified `BlenderSubmitter` (18526a8): `_create_bundle_internal` adapts the UI settings via `_ui_settings_to_submitter_settings` and calls `submitter.get_job_template` / `get_parameter_values`, mirroring deadline-cloud-for-maya #435. Follow-up (249a37d) also extracted the shared scene-gathering helpers into `blender_utils` (`get_render_engine`, `resolve_output_path`, `resolve_output_file_prefix`, `classify_auto_detected_paths`, `resolve_gpu_settings`) so both flows call one implementation and the earlier drift (output-path fallbacks, OCIO dirs, GPU, cameras) is gone.

leon-li-inspire: Agreed with the model direction, and I've done step 1 of your proposal here: the scene-gathering logic is now shared (in `blender_utils`/`ocio_utils`) and called from both the GUI and the unified `BlenderSubmitter`, and the GUI delegates its bundle build to `BlenderSubmitter`. The full consolidation — collapse `BlenderSubmitterUISettings`/`BlenderSubmitterSettings` into one `from_scene()` model and delete `_to_native_settings` — is the tracked follow-up so this PR isn't blocked on it. One open question: should the single-model pattern be the template for all DCCs (so Maya/Nuke/Houdini adopt it too)? Happy to align on that before doing the follow-up, since it affects cross-DCC consistency.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
leon-li-inspire: `_to_native_settings` is still needed today, and `get_parameter_values` can't just cast: `template_filling.fill_job_template`/`get_parameter_values` read fields that live only on `BlenderSubmitterUISettings`, not on `BlenderSubmitterSettings` — e.g. `name` (vs `job_name`), `ui_group_label`, `image_resolution` (a `"W, H"` string vs our int `image_width`/`image_height`), and the `*_parameter_name` fields. A cast would `AttributeError` at runtime on those. So there are two adapters by design right now: `_ui_settings_to_submitter_settings` (UI → engine settings, GUI boundary) and `_to_native_settings` (engine settings → the UISettings shape `template_filling` consumes). Both collapse in the model follow-up (one `from_scene()` model that `template_filling` reads directly) — I kept them for this PR to bound the diff. Want me to fold the retarget into this PR instead?

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,280 @@
...
+from typing import TYPE_CHECKING, Any, Optional, cast
+
+# BaseSubmitter/BaseSubmitterSettings live in deadline-cloud #1245, which is not
+# yet in a published `deadline` release; until it ships, mypy resolves against
+# the base-less PyPI package and flags attr-defined. Suppress that transient
+# error (it clears automatically once the base is released).
+from deadline.client.api import (  # type: ignore[attr-defined]
+    BaseSubmitter,
+    BaseSubmitterSettings,
+)
```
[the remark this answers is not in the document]
leon-li-inspire: Correct — this is a runtime `ImportError`, not just a mypy issue; it's the same reason the CI "Run Tests"/import jobs are red. `BaseSubmitter`/`BaseSubmitterSettings` land in deadline-cloud #1245, which isn't in a published `deadline` release yet. Plan: once #1245 ships, pin `pyproject.toml` to the minimum `deadline` version that exports them (and drop the `# type: ignore`). This PR depends on #1245 and shouldn't merge before it — same posture as deadline-cloud-for-maya #435. I'll update the pin in the same change that bumps the dependency; flagging here so it's not lost.

test/unit/deadline_submitter_for_blender/test_submitter.py
leon-li-inspire: Done in 503be32 — renamed to `test_gui_submitter.py`.

test/unit/deadline_submitter_for_blender/test_submitter_engine.py
leon-li-inspire: Done in 503be32 — `test_submitter_engine.py` renamed to `test_submitter.py` (and the former `test_submitter.py` moved to `test_gui_submitter.py`).

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,243 @@
...
+from deadline.client.api import BaseSubmitter, BaseSubmitterSettings
+
+import bpy  # type: ignore[import]
+
+from deadline.client.job_bundle.submission import AssetReferences
+
+if TYPE_CHECKING:
+    # Only referenced in a string cast() annotation below; import under
+    # TYPE_CHECKING so it is not an unused runtime import (CodeQL).
+    from deadline.client.job_bundle.parameters import JobParameter
```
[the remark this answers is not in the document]
leon-li-inspire: `JobParameter` is only referenced in a string `cast("list[JobParameter]", …)` annotation, so it's imported under `TYPE_CHECKING` (no runtime import). This CodeQL alert is no longer in the open set on the current head. It's still needed for the mypy/static typing of the queue-parameters cast, so it stays under `TYPE_CHECKING` rather than being removed.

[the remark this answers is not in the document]
leon-li-inspire: `JobParameter` is only referenced in a string `cast("list[JobParameter]", …)` annotation, so it's imported under `TYPE_CHECKING` (no runtime import). This CodeQL alert is no longer in the open set on the current head. It's still needed for the mypy/static typing of the queue-parameters cast, so it stays under `TYPE_CHECKING` rather than being removed.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
leon-li-inspire: Done in 320596e — `_to_native_settings` is removed. `template_filling`'s builders now annotate and read `BlenderSubmitterSettings` directly (only change was `settings.name` → `settings.job_name`; everything else they read off `settings` already existed on `BlenderSubmitterSettings`, and the `*_parameter_name`/`image_resolution` values come from `CommonLayerSettings`, not `settings`). `get_job_template`/`get_parameter_values` now just `cast(...)` the settings and pass them straight through — no round-trip. The `BlenderSubmitterSettings` import in `template_filling` is under `TYPE_CHECKING` to avoid a circular import (submitter imports the builders). This is the retarget you suggested; the full single-`from_scene()`-model consolidation remains the follow-up.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
```diff
@@ -210,6 +197,47 @@ def _create_bundle(
...
+    ``_to_native_settings``. This is the single adaptation point; both the GUI
+    and the unified API path then run the same BlenderSubmitter builders.
+
+    The effective frame range is resolved here (override -> the user's range,
+    otherwise the scene range) so the engine sees a ready-to-use ``frame_list``.
+    """
+    frame_list = settings.frame_list if settings.override_frame_range else bu.get_frames()
+    width, height = bpy.context.scene.render.resolution_x, bpy.context.scene.render.resolution_y
+
+    return BlenderSubmitterSettings(
```
moonyuet, line 217: ~This is errored out for now as the relevant dataclass doesn't have that much arguments.~        ~You might need to add the attributes back~
~Same thing happened to the relevant PRs in Maya.~
I just find out the related attributes have been implemented in the BaseSubmitterSettings.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,252 @@
...
+            output_file_prefix = settings.output_file_prefix
+            image_resolution = (settings.image_width, settings.image_height)
+
+        # Populate the renderable cameras so the "All Renderable Cameras"
+        # selection produces a real Camera task dimension (mirrors the dialog).
+        # Best-effort: an invalid scene name must not block template building.
+        try:
+            renderable_camera_names = get_renderable_cameras(scene_name)
+        except Exception:  # noqa: BLE001
+            renderable_camera_names = []
```
moonyuet: It will return empty list if you dont have any rendeable cameras in blender. See the test below:
Image
```suggestion
        renderable_camera_names = get_renderable_cameras(scene_name)
```

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,252 @@
...
+        except Exception:  # noqa: BLE001
+            # Auto-detection is best-effort: never let a scan failure block the
+            # submission (the .blend + explicit directories are still attached).
+            pass
+
+        # Attach the directories referenced by the OCIO config (shared with the
+        # dialog flow) so custom color management resolves on the farm. Honor an
+        # explicitly-set ocio_config_path, else fall back to the active config.
+        try:
+            ocio_config_path = getattr(settings, "ocio_config_path", "")
```
moonyuet: I assume there is always ocio_config_path but it is only that it would return empty string if there is no value filled?(knowing that the default is empty string not None?)

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,252 @@
...
+            input_directories.update(ocio.get_current_ocio_referenced_dirs(ocio_config_path))
+        except Exception:  # noqa: BLE001
+            # Best-effort, as above: a bad/unreadable OCIO config must not
+            # block submission.
+            pass
```
moonyuet:
```suggestion

        referenced_dirs = ocio.get_current_ocio_referenced_dirs(settings.ocio_config_path)
        input_directories.update(set(referenced_dirs))
```

BigRoy: This should really not be a bare except that is silently ignored, right? Agree with @moonyuet here.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
BigRoy: Bare exceptions here will just obfuscate/hide actual errors.

In this case - the most likely error here would be that the `scene_name` passed is not in `bpy.data.scenes`.

Also, this logic seems a bit off: https://github.com/leon-li-inspire/deadline-cloud-for-blender/blob/898aa5f7f0c1e85973b1a25151306efd82b0326e/src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/blender_utils.py#L22C18-L23

Why would you do:
```
scene_name = bpy.data.scenes[saved_scene_name].name
```
To then access the layers by
```
bpy.data.scenes[scene_name].view_layers
```
Isn't that equivalent to:
```
bpy.data.scenes[saved_scene_name].view_layers
```
?

leon-li-inspire: Addressed in \`0ece322\`.

- Removed the \`try/except → []\`: \`_build_common_layer_settings\` now calls \`get_renderable_cameras(scene_name)\` directly, so a scene with no renderable cameras returns \`[]\` naturally, and an invalid \`scene_name\` surfaces as a \`KeyError\` (same as \`_resolve_view_layer_names\`) instead of being silently swallowed. @BigRoy — agreed on the bare-except point.
- Good catch on the redundant round-trip: \`get_renderable_cameras\`/\`get_renderable_view_layers\` now index \`bpy.data.scenes[saved_scene_name]\` directly instead of \`scenes[scenes[name].name]\`. They are equivalent, so I simplified to the direct form.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
leon-li-inspire: Correct — \`ocio_config_path\` is a real field on \`BlenderSubmitterSettings\` with default \`""\` (never \`None\`). \`get_settings()\` sets it from \`ocio.get_ocio_path()\`, which returns \`""\` when \`$OCIO\` is unset. Downstream, \`get_current_ocio_referenced_dirs("")\` falls back to the active config and returns \`[]\` when none is set. In \`0ece322\` I dropped the now-redundant \`getattr(settings, "ocio_config_path", "")\` and read the field directly.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
leon-li-inspire: Addressed in \`0ece322\`. Adopted the two-line form and, per @BigRoy, replaced the bare \`except: pass\` with log-and-continue: a malformed OCIO config now logs a warning (with \`exc_info\`) and submission proceeds without the OCIO dirs, rather than silently swallowing. (Kept the guard rather than letting it hard-fail, since a bad OCIO config should not block an otherwise-valid submission — but it is no longer invisible.)

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,273 @@
...
+        # default camera (mirrors the dialog) when none is resolved.
+        if scene.camera:
+            settings.camera_selection = scene.camera.name
+        else:
+            settings.camera_selection = "Use Default Camera"
+
+        # Read GPU settings from the scene/preferences (shared helper), so API
+        # submissions honor the user's GPU choice and default to the "NONE"
+        # sentinel (not "None") that the template/adaptor expect.
+        settings.enable_gpu, settings.gpu_device = resolve_gpu_settings()
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in \`0ece322\`. Folded the gating into the shared \`resolve_gpu_settings()\` helper: \`gpu_device\` is only set to \`compute_device_type\` when \`enable_gpu\` (i.e. \`scene.cycles.device == "GPU"\`); otherwise it stays \`"NONE"\`. This matches \`scene_settings_widget.update_settings\` exactly, so a CPU scene on a workstation with a GPU compute device selected no longer renders on GPU via the API path. Both flows now agree. Added a unit test (\`test_cpu_scene_with_configured_gpu_device_stays_none\`).

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/blender_utils.py
```diff
@@ -53,18 +53,128 @@ def get_scene_resolution(saved_scene_name):
 
 
 def get_scene_name() -> str:
-    """Construct and return a name for the current scene based on the currently opened `.blend` file."""
+    """Construct and return a name for the current scene based on the currently opened `.blend` file.
+
+    The dialog submitter uses this as the default *job* name
+    (``open_deadline_cloud_dialog.py``), so it must stay derived from the
+    ``.blend`` file name. Callers that need the active scene's name as a
+    ``bpy.data.scenes`` key (e.g. the unified ``BlenderSubmitter`` path) must
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in \`0ece322\`. \`get_render_engine()\` now raises \`DeadlineOperationError\` for any engine outside the template's \`allowedValues\` (Cycles/EEVEE/Workbench) instead of forwarding \`engine.lower()\`. A third-party active engine (V-Ray, Octane, Redshift, …) now fails early with "Render engine 'X' is not supported. Switch … to Cycles, EEVEE, or Workbench" rather than surfacing as an opaque CreateJob ValidationException. Test updated (\`test_unsupported_engine_raises\`).

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/ocio_utils.py
```diff
@@ -30,3 +30,19 @@ def get_ocio_referenced_dirs(config: ocio.Config) -> list[str]:
...
+    the ``OCIO`` env var when not given) and returns the directories it
+    references, so they can be added as job input attachments. Empty when no
+    OCIO config is set. Shared by the GUI submitter and the unified API path so
+    both attach the same OCIO-referenced directories.
+    """
+    ocio_path = ocio_path or get_ocio_path()
+    if not ocio_path:
+        return []
+    config = get_ocio_config(ocio_path)
+    return [str(path) for path in get_ocio_referenced_dirs(config)]
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in \`0ece322\`. \`get_ocio_referenced_dirs\` now takes the config path and resolves relative \`search_path\` entries against that config's own directory (falling back to \`$OCIO\` when not given), and \`get_current_ocio_referenced_dirs\` threads the resolved \`ocio_path\` through. So an explicit \`ocio_config_path\` that differs from \`$OCIO\` now resolves its relative search paths against the correct base directory.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,243 @@
...
+
+    renderer_name: str = "cycles"
+    scene_name: str = "Scene"
+    view_layer_selection: str = ""
+    camera_selection: str = ""
+    image_width: int = 1920
+    image_height: int = 1080
+    output_file_prefix: str = "output_####"
+    enable_gpu: bool = False
+    gpu_device: str = "None"
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed. \`get_settings()\` now populates \`enable_gpu\`/\`gpu_device\` from the scene via \`resolve_gpu_settings()\`, and the default is the \`"NONE"\` sentinel (uppercase) matching the template/adaptor. The subsequent enable_gpu gating (this thread's sibling) is handled in \`0ece322\`.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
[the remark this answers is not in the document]
leon-li-inspire: Done. deadline-cloud \`0.60.2\` (which exports \`BaseSubmitter\`/\`BaseSubmitterSettings\`) has shipped, and \`pyproject.toml\` is pinned to \`deadline >= 0.60.2, < 0.61\` (both base and \`[gui]\`) in \`8498c46\`. Verified 0.60.0/0.60.1 do NOT export the symbols, so 0.60.2 is the correct floor. The \`# type: ignore[attr-defined]\` was also dropped since the symbols now resolve.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,252 @@
...
+    get_frames,
+    get_render_engine,
+    get_renderable_cameras,
+    get_renderable_view_layers,
+    resolve_gpu_settings,
+    resolve_output_file_prefix,
+    resolve_output_path,
+)
+from .template_filling import (
+    CommonLayerSettings,
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in \`531529d\`. \`BlenderSubmitterSettings\` moved into \`template_filling\` (next to its sibling dataclasses the builders consume) and re-exported from \`submitter\` via \`__all__\`, so the dependency now flows one way (\`submitter -> template_filling\`) with no back-import. Verified locally with CodeQL 2.26.1 running the same \`ModuleLevelCyclicImport.ql\`/\`UnusedImport.ql\` queries: 5 alerts before, 0 after.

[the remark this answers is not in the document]
leon-li-inspire: Resolved in \`531529d\`: \`BlenderSubmitterSettings\` moved into \`template_filling\` and re-exported from \`submitter\`, so the dependency flows one way (\`submitter -> template_filling\`) with no back-import. Confirmed on the current head (\`70a6a7d\`): the CodeQL check passes, there are no open \`py/unsafe-cyclic-import\` alerts on the PR ref, and re-running \`ModuleLevelCyclicImport.ql\` locally reports 0. These 4 annotations are stale (dated 2026-07-16, from the pre-fix state) and did not auto-close; resolving.

leongdl (dismissed):
i think it is ok, maybe a few nits.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
```diff
@@ -75,26 +75,14 @@ def create_deadline_dialog(parent=None) -> SubmitJobToDeadlineDialog:
...
-
-    if bpy.context.preferences.addons["cycles"].preferences.compute_device_type != "NONE":
-        settings.gpu_device = bpy.context.preferences.addons[
-            "cycles"
-        ].preferences.compute_device_type
+    # Read GPU settings from the scene/preferences (shared helper).
+    settings.enable_gpu, settings.gpu_device = bu.resolve_gpu_settings()
 
     # Load and set sticky settings, if any.
     settings.load_sticky_settings(settings.project_path)
```
leongdl, line 88: Nit (but it can silently undo the `"NONE"` sentinel fix): `gpu_device` is a **sticky** field on `BlenderSubmitterUISettings`, and its default there is still `"None"` (mixed case). Since `load_sticky_settings()` runs *after* `resolve_gpu_settings()` sets the fresh `"NONE"` value, any user with an existing `.deadline_render_settings.json` written by the old addon (which saved `"None"` for CPU scenes) will have the stale `"None"` loaded back over the corrected value — flowing through `_ui_settings_to_submitter_settings` into the bundle and re-triggering exactly the adaptor GPU-path bug this PR fixes.

Suggest: change the `BlenderSubmitterUISettings.gpu_device` default to `"NONE"` and normalize the value on sticky load (e.g. map legacy `"None"` → `"NONE"`), so existing sticky files can't resurrect the old behavior.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,236 @@
...
+        # default camera (mirrors the dialog) when none is resolved.
+        if scene.camera:
+            settings.camera_selection = scene.camera.name
+        else:
+            settings.camera_selection = "Use Default Camera"
+
+        # Read GPU settings from the scene/preferences (shared helper), so API
+        # submissions honor the user's GPU choice and default to the "NONE"
+        # sentinel (not "None") that the template/adaptor expect.
+        settings.enable_gpu, settings.gpu_device = resolve_gpu_settings()
```
leongdl, line 91: Nit: `enable_gpu` is a write-only field — `get_settings()` populates it and the dialog adapter maps it, but nothing in `template_filling` ever reads it; only `gpu_device` reaches the bundle. An API caller who sets `enable_gpu=True` but leaves `gpu_device="NONE"` gets a CPU render with no warning.

Either have the builders derive the gating from it (e.g. warn/raise when `enable_gpu=True` and `gpu_device="NONE"`), or drop it from `BlenderSubmitterSettings` to avoid implying it has an effect.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
```diff
@@ -0,0 +1,236 @@
...
+    def _resolve_view_layer_names(self, settings: BaseSubmitterSettings) -> list[str]:
+        """Return the view layers to render, honoring view_layer_selection.
+
+        An empty selection (or the "All Renderable Layers" sentinel) renders
+        every renderable view layer; any other value renders that single layer.
+        Mirrors the dialog flow so the GUI and unified API paths agree.
+        """
+        # scene_name must be a key in bpy.data.scenes (settings.job_name is the
+        # .blend file name, which is not, and would raise KeyError downstream).
+        scene_name = getattr(settings, "scene_name", None) or settings.job_name
```
leongdl: Nit: this fallback to `settings.job_name` guarantees the KeyError it's trying to avoid — the comment two lines up correctly says `job_name` is the .blend filename and *not* a `bpy.data.scenes` key. When `scene_name` is empty, falling back to `job_name` just produces a more confusing `KeyError` deep in `get_renderable_view_layers`. Failing fast on an empty `scene_name` (or raising `DeadlineOperationError` with an actionable message, like `get_render_engine` does) would be clearer for headless callers.

leongdl (commented):
Compared the refactored flow against the old UISettings-driven path. Good news first: the hardcoded parameter names in `_build_common_layer_settings` (`Frames`, `OutputFileName`, `ResolutionX`, `ResolutionY`, `Blender Settings`) all match the old `BlenderSubmitterUISettings` defaults exactly, so the generated templates are unchanged. No user-facing feature appears dropped by the delegation refactor.

A few nits found in the comparison (inline comments below), plus two non-blocking notes:

- The `*_parameter_name` / `ui_group_label` fields on `BlenderSubmitterUISettings` are now dead in the GUI path (set but never read, since `_build_common_layer_settings` hardcodes them). Worth removing in a follow-up to avoid confusion. `scene_name_parameter_name` (sticky) appears to have been unused even before this PR — same follow-up.
- The API path intentionally has no equivalent of `run_sanity_checks` (unsaved .blend → empty `project_path`/`OutputDir` only fails later at CreateJob). If that's by design for headless callers, fine — might be worth a docstring note on `get_settings()`.

AlexTranAmz (dismissed):

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/open_deadline_cloud_dialog.py
leon-li-inspire: Fixed in \`22d5303\`. Changed the \`BlenderSubmitterUISettings.gpu_device\` default to \`"NONE"\`, and \`load_sticky_settings()\` now normalizes a legacy \`"None"\` → \`"NONE"\` after loading — so an old \`.deadline_render_settings.json\` can no longer resurrect the mixed-case value over the corrected one. Added \`test_load_sticky_settings_normalizes_legacy_none_gpu_device\` + \`test_ui_settings_gpu_device_defaults_to_none_sentinel\`.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
leon-li-inspire: Fixed in \`22d5303\`. \`_resolve_view_layer_names\` now raises \`DeadlineOperationError\` on an empty \`scene_name\` with an actionable message (mirroring \`get_render_engine\`), instead of falling back to \`job_name\`. Added \`test_resolve_view_layers_empty_scene_name_raises\`.

src/deadline/blender_submitter/addons/deadline_cloud_blender_submitter/submitter.py
leon-li-inspire: Good point — \`enable_gpu\` is indeed write-only today (only \`gpu_device\` reaches the bundle). This is a small design choice between two options, so flagging rather than picking unilaterally:

1. **Keep + gate**: have \`get_settings\`/the builders derive \`gpu_device="NONE"\` when \`enable_gpu\` is false (already the effective behavior via \`resolve_gpu_settings\`), and warn if an API caller sets \`enable_gpu=True\` with \`gpu_device="NONE"\`. Keeps the field meaningful for callers who think in terms of "GPU on/off".
2. **Drop \`enable_gpu\`** from \`BlenderSubmitterSettings\` so \`gpu_device\` (\`"NONE"\` vs a device type) is the single source of truth — simpler, but changes the public settings shape the AYON consumer imports.

Leaning toward (1). Will follow up in a separate commit once the direction is confirmed.

moonyuet:
I tested with the latest change in my side, published and rendered successfully.
image
image

leongdl (approved):

AlexTranAmz (approved):

PR #382, opened 2026-08-06
chore(deps): bump pypa/gh-action-pypi-publish, ruff, and twine

leongdl:
Combined dependency upgrades from 3 dependabot PRs:

- **pypa/gh-action-pypi-publish**: 1.14.1 → 1.14.2
- **ruff**: 0.15.* → 0.16.*
- **twine**: 6.* → 7.* (Python 3.10+), keep 6.* for Python 3.9 (twine 7 dropped 3.9 support)

### Ruff 0.16 changes

Ruff 0.16 enables ~30 new lint rules by default that weren't enforced in 0.15. This triggered 161 errors across the codebase. We handled them in three tiers:

**Fixed (18 rules):** Auto-fixable rules with trivial, safe changes — unused noqa comments, collapsible if-statements, f-string modernization, unnecessary literals, shebang issues, and `subprocess.run` without explicit `check` argument.

**Suppressed (12 rules):** These fall into two categories:

1. **Python 3.9 compatibility (4 rules):** `UP006`, `UP035`, `UP045`, `SIM117` — these want modern syntax (`dict` instead of `typing.Dict`, `X | None` instead of `Optional[X]`, parenthesized context managers) that either requires `from __future__ import annotations` in every file or Python 3.10+. Since we support 3.9, fixing these would require adding `__future__` imports to 47+ files for no functional benefit.

2. **Intentional patterns or high-churn rules (8 rules):**
   - `FA100` (47 files) / `I001` (36 files): Would touch nearly every file just to add imports or reorder them — too noisy for a dep bump PR.
   - `BLE001`: Broad `except Exception` catches are intentional in entry points and error boundaries.
   - `TRY002`: `raise Exception(...)` in scripts is deliberate — creating custom exceptions for one-off scripts is over-engineering.
   - `B008`, `PLR1704`, `RUF012`, `SIM113`: 1-2 occurrences each with intentional patterns (Blender API conventions, manual iteration logic).

These suppressed rules can be addressed incrementally in follow-up PRs if desired.

Closes #377, closes #379, closes #381

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,20 +113,19 @@ exclude = [".kiro"]
 [tool.ruff.lint]
 ignore = [
     "E501",
-    # Rules newly enforced in ruff 0.16 - suppress to keep upgrade minimal
+    # Rules newly enforced in ruff 0.16 - suppress for Python 3.9 compat or intentional patterns
     "B008",      # function-call-in-default-argument
-    "BLE001",    # blind-except
-    "FA100",     # future-annotations-import
-    "I001",      # unsorted-imports
+    "BLE001",    # blind-except (intentional broad catch)
+    "FA100",     # future-annotations-import (47 files)
+    "I001",      # unsorted-imports (36 files - conflicts with isort config)
     "PLR1704",   # redefined-argument-from-local
-    "PLW1510",   # subprocess-run-without-check
     "RUF012",    # mutable-class-default
     "SIM113",    # enumerate-for-loop
-    "SIM117",    # multiple-with-statements
-    "TRY002",    # raise-vanilla-class
-    "UP006",     # non-pep585-annotation
-    "UP035",     # deprecated-import
-    "UP045",     # non-pep604-annotation
+    "SIM117",    # multiple-with-statements (Python 3.9 compat)
+    "TRY002",    # raise-vanilla-class (scripts)
+    "UP006",     # non-pep585-annotation (Python 3.9 compat)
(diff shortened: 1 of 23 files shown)


pyproject.toml
```diff
@@ -108,9 +108,44 @@ module = ["bpy.*", "qtpy.*", "PyOpenColorIO.*"]
 
 [tool.ruff]
 line-length = 100
+exclude = [".kiro"]
 
 [tool.ruff.lint]
-ignore = ["E501"]
+ignore = [
+    "E501",
+    # Rules newly enforced in ruff 0.16 - suppress to keep upgrade minimal
```
leongdl: We have to support Python 3.9, and this led to significant rewrite for no reason.

requirements-testing.txt
```diff
@@ -8,6 +8,7 @@ pytest == 9.*; python_version >= "3.10"
 pytest == 8.*; python_version < "3.10"
 pytest-cov == 7.*
 pytest-xdist == 3.*
-ruff == 0.15.*
-twine == 6.*
+ruff == 0.16.*
+twine == 7.*; python_version >= "3.10"
+twine == 6.*; python_version < "3.10"
```
epmog: pretty sure this is only used for publishing which we have set to 3.13. Perhaps it should be split out so we don't need this. low-risk either way

epmog (approved):

AlexTranAmz (approved):

Issue #383, opened 2026-08-11
Bug: Blender 5.2 - Python313 compatability issue on Windows


patrickjahns:
### Describe the bug

When running deadline-cloud with Blender 5.2 we encountered the issue, that the BlenderClient always exits with the error message: 
```
OSError: BlenderClient cannot connect to the Adaptor because the server at the path defined by the environment variable BLENDER_ADAPTOR_SERVER_PATH does not exist. Got: \\.\pipe\AdaptorServerNamedPipe_8788
```

When investigating this issue, we found out, that the problem resides to how `os.path.exists()` resolve on Python311 and Python313 on windows. Namely the way it handles the check how deadline-cloud / the openjd-adapter use it:

> Python 3.11: os.stat() on Windows relied on FindFirstFileW as a fallback mechanism for device namespace paths. FindFirstFileW performed a directory catalog lookup in the Named Pipe File System (NPFS) without opening a handle or requiring an available instance slot, bypassing connection limits.

> Python 3.13: CPython updated os.stat() to use handle-based queries via CreateFileW(..., FILE_READ_ATTRIBUTES, ...) instead of FindFirstFileW. Because a busy single-instance pipe cannot accept new attribute handles, CreateFileW fails with ERROR_PIPE_BUSY (231), causing os.stat() to fail and os.path.exists() to evaluate to False.

**full log**
```
2026/08/10 21:55:24+02:00 INFO: Connected successfully
2026/08/10 21:55:24+02:00 openjd_env: OPENJD_ADAPTOR_SOCKET=\\.\pipe\AdaptorNamedPipe_8788
2026/08/10 21:55:24+02:00 ADAPTOR_OUTPUT: INFO: Running in background daemon mode.
2026/08/10 21:55:24+02:00 ADAPTOR_OUTPUT: INFO: Creating Named Pipe with name: \\.\pipe\AdaptorNamedPipe_8788
2026/08/10 21:55:24+02:00 ADAPTOR_OUTPUT: INFO: Loading 'init_data' schema from C:\Python313\Lib\site-packages\deadline\blender_adaptor\BlenderAdaptor\schemas\init_data.schema.json
2026/08/10 21:55:24+02:00 ADAPTOR_OUTPUT: INFO: Loading 'run_data' schema from C:\Python313\Lib\site-packages\deadline\blender_adaptor\BlenderAdaptor\schemas\run_data.schema.json
[...]

PR #384, opened 2026-08-11
fix: natively check for namedpip existance on windows

patrickjahns:
Blender 5.2 switched to Python313 and with that switch the BlenderClient always exited with an Error

```
OSError: BlenderClient cannot connect to the Adaptor because the server at the path defined by the environment variable BLENDER_ADAPTOR_SERVER_PATH does not exist. Got: \\.\pipe\AdaptorServerNamedPipe_8788
```

See #383 for complete background. In summary - CPython changed the behaviour for `os.stat.exists()` between 311 and 313. 

**background**

> Why os.path.exists() Always Fails on Python 3.13 for Windows Pipes
> - CPython 3.13 Refactoring: In Python 3.13, CPython optimized os.stat() / os.path.exists() on Windows to use the GetFileAttributesExW Win32 API call.
> - NPFS Incompatibility: The Windows Named Pipe File System (NPFS / \\.\pipe\) does not support file attribute queries. When GetFileAttributesExW is called on a named pipe, Windows returns ERROR_FILE_NOT_FOUND (WinError 2).
> - The Result: Because GetFileAttributesExW fails, Python 3.13's os.path.exists(r"\\.\pipe\...") always evaluates to False, regardless of how long you wait or whether the pipe is active and healthy.
> The previous check worked in Python 3.10/3.11 because older CPython versions fell back to CreateFileW inside os.stat().

- Implemented a helper method to check on windows explicitely if we are dealing with a pipe

- Blender 5.2 will work

- Locally 

```
to be run
```

- No change

- No

No
----

### DIFF
--- a/src/deadline/blender_adaptor/BlenderClient/blender_client.py
+++ b/src/deadline/blender_adaptor/BlenderClient/blender_client.py
@@ -3,6 +3,8 @@
 from __future__ import annotations
 
 import os
+import ctypes
+import sys
 from types import FrameType
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


leongdl:
Please sign the commit with `-s`

leongdl (approved):

leongdl (changes requested):
Please commit with -s

leongdl (commented):
Please re-submit with commit --amend -s, and also fix the linting error:
```
 class BlenderClient(ClientInterface):
     def __init__(self, server_path: str) -> None:
         super().__init__(server_path=server_path)
         print(f"BlenderClient: Blender Version {bpy.app.version_string}")

Oh no! 💥 💔 💥
1 file would be reformatted, 58 files would be left unchanged.
```

hatch run fmt all should do it.

patrickjahns:
> Please re-submit with commit --amend -s, and also fix the linting error:
> 
> ```
>  class BlenderClient(ClientInterface):
>      def __init__(self, server_path: str) -> None:
>          super().__init__(server_path=server_path)
>          print(f"BlenderClient: Blender Version {bpy.app.version_string}")
> 
> Oh no! 💥 💔 💥
> 1 file would be reformatted, 58 files would be left unchanged.
> ```
> 
> hatch run fmt all should do it.



Done - let me know if the other 🤖  remarks should be addressed 👍

leongdl (approved):
Thank you for the contribution.

leongdl (approved):

leongdl:
Looking for a second reviewer from my team to close on the 5.2 Blender support. We will release this as part of our next update.

seant-aws (approved):
thanks for the change ! :)

PR #392, opened 2026-09-18
fix: require deadline 0.60.4 with the console extra and ship a loadable awscrt

andychoquette:
AWS Console sign-in fails to authenticate from the submitter. The `deadline` floors (0.60.2) sit below 0.60.4 where the `console` extra first exists, so pip silently backtracks and drops awscrt; and the flat deps bundle ships a single `_awscrt.abi3.so` chosen by the build host instead of the lowest supported Python, which is the only copy every interpreter can load (abi3 is forward-compatible only).

- Bump both `deadline` floors to `>= 0.60.4,< 0.61`; gui extra now requests `deadline[gui,console]` (console pulls awscrt, which botocore's LoginProvider needs to sign DPoP proofs in-process)
- Console extra stays off base dependencies: those resolve into the adaptor artifact under `--only-binary --platform` tags with no suitable awscrt wheel
- `depsBundle.py` resolves the console extra into the bundle
- awscrt and pyyaml added to `NATIVE_DEPENDENCIES` (fetched per supported Python)
- Native trees merge in ascending Python version order, first tree wins a filename collision and overwrites the base environment
- Unit tests guard the merge selection, download order, pip list casing, and the declared floors/extras

Ports aws-deadline/deadline-cloud-for-cinema-4d#525 and aws-deadline/deadline-cloud-for-cinema-4d#554 to this repo.


test/unit/deadline_submitter_for_blender/test_deps_bundle_native_merge.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def _write(path: Path, content: str) -> None:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(content)
+
+
+@pytest.fixture
+def supported_versions() -> list[str]:
+    versions = sorted(depsBundle.SUPPORTED_PYTHON_VERSIONS, key=_version_key)
```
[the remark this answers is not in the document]
andychoquette: Both correct — fixed in 4cac104, matching how cinema-4d resolved the same finding (its PR #554, commit 05b8c1f):

- `AWSCRT_LAST_NON_ABI3_VERSION = "3.10"` names the boundary explicitly. The fixture assigns version-specific vs abi3 filenames by comparing each version to the boundary rather than by index, and `lowest_abi3_version` is computed as `min(v for v in supported_versions if v > boundary)`.
- Added a fixture guard asserting supported versions exist on both sides of the boundary. This subsumes the `>= 2` vacuous-pass concern: with only one version on either side, the guard fails loudly instead of the collision assertion passing vacuously.

Kept consistent with cinema-4d for the shared-bundler consolidation.

scripts/depsBundle.py
```diff
@@ -13,7 +13,18 @@
...
+#
+# awscrt is here because its wheels are not uniformly abi3: Python 3.10 gets
+# _awscrt.cpython-310-<platform>.so while 3.11+ get _awscrt.abi3.so. Resolving it only in
+# the base environment would ship whichever the build host produced, so any Blender whose
+# interpreter that single artifact does not cover would fail to import awscrt and AWS
+# Console sign-in would break there.
+# pyyaml is here because it ships a version-specific `_yaml` extension module: resolved only
+# in the base environment it lands built for a single interpreter, and pyyaml hides that by
+# falling back to its pure-Python parser on the others.
+NATIVE_DEPENDENCIES = ["xxhash", "psutil", "awscrt", "pyyaml"]
```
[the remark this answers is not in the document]
andychoquette: Added "3.9" to SUPPORTED_PYTHON_VERSIONS in 4cac104 — confirmed empty by pip, not just by intent.

Checked cp39 wheel availability for linux/win/mac for the exact versions this bundle resolves:

- awscrt at 0.31.2 (the exact version botocore[crt]>=1.42.89 pins, which is what `deadline[console]>=0.60.4` resolves to): cp39 wheels exist for manylinux2014_x86_64, win_amd64, and macosx_10_15_universal2 (the modern mac tag pip actually resolves for this script, since depsBundle.py never passes `--platform`). No cp39 wheel exists for the legacy macosx_10_9_x86_64 tag at 0.31.2, but that tag is only used by `create_adaptor_packaging_artifact.sh` for the adaptor, not by this bundle script.
- pyyaml (6.0.3) and xxhash (4.0.1): cp39 wheels exist for all three platforms.
- psutil ships cp36-abi3, so it already loaded on 3.9 regardless.

Verified via `pip download --no-deps --only-binary=:all: --python-version 3.9 --platform <tag> <pkg>==<version>` for each.

The abi3 winner stays the 3.11 build — 3.9 and 3.10 both get version-specific (non-abi3) awscrt artifacts now, same shape as before, just two version-specific writers instead of one. Updated the merge-test fixture/assertions in test_deps_bundle_native_merge.py accordingly (see reply on that comment) and extended `test_version_specific_artifacts_are_kept_for_every_supported_version` to check every version below the boundary, not just the lowest.

scripts/depsBundle.py
```diff
@@ -179,28 +193,57 @@ def _get_package_version(package: str, install_path: Path) -> str:
...
+    # The bundle is the submitter, which needs AWS Console sign-in. The console extra is
+    # requested here rather than declared in project.dependencies, because those are also
+    # resolved into the adaptor package, where a compiled awscrt wheel is both unusable and
+    # unavailable for one of the platform tags that build targets (see pyproject.toml).
+    #
+    # Requesting the extra rather than installing awscrt directly means the bundle tracks
+    # whatever the extra actually requires -- notably a botocore floor, since the console
+    # login provider lives in botocore, not in deadline -- and takes awscrt from the exact
+    # version botocore's crt extra pins, rather than resolving it independently and drifting.
+    dependencies_for_pip = [_add_console_extra(dep) for dep in dependencies]
```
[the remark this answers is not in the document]
andychoquette: Fixed in 4cac104, matching how nuke resolved the same finding (its PR #349, commit e556c0b), plus the stronger assertion suggested:

- Parametrized `test_add_console_extra_pins_behavior`: plain requirement gains `[console]` preserving the specifier, an existing extra is preserved (not replaced), the result is idempotent, and non-deadline requirements pass through untouched.
- `test_build_base_environment_requests_the_console_extra`: monkeypatches `subprocess.run` and asserts the pip argv `_build_base_environment` builds contains a `deadline[...console...]` requirement.
- `test_native_dependencies_include_awscrt_and_pyyaml`: pins both in NATIVE_DEPENDENCIES.
- `test_add_console_extra_changes_the_real_base_dependencies`: the "stronger still" case — runs `_add_console_extra` over the actual `pyproject.toml` dependencies (via `depsBundle._get_dependencies`) and asserts at least one entry changes, so a rename of the `deadline` requirement fails this test instead of silently shipping a bundle with no awscrt.

Left `test_gui_extra_requests_the_console_extra` as-is — noting here per your comment that it guards the `pip install deadline-cloud-for-blender[gui]` declaration path, not the bundle build; the new tests above cover the bundle path directly.

test/unit/deadline_submitter_for_blender/test_deps_bundle_native_merge.py
```diff
@@ -0,0 +1,194 @@
...
+        ),
+        key=_version_key,
+    )
+    shipped = (merged_bundle / ABI3_ARTIFACT).read_text()
+
+    assert shipped == lowest_abi3_version, (
+        f"{ABI3_ARTIFACT} was built for Python {shipped}, so it cannot be imported by "
+        f"Python {lowest_abi3_version}; the copy built for the lowest supported abi3 "
+        f"version is the one every supported interpreter can load"
+    )
```
phil-IO-p: The fixture seeds the base environment with `supported_versions[-1]` at `:81`, and the assertion at `:126` `lowest_abi3_version`. With today's list those are `3.13` and `3.11` so it bites, but the guard at `:84-86` only requires one supported version on each side of the boundary. If `SUPPORTED_PYTHON_VERSIONS` ever holds exactly one version above 3.10, the seeded value and the expected value become the same string and the assertion passes even if `_copy_native_to_base_env` copies nothing.                                                                      
                                                                              
Would seeding a sentinel instead work, something like `_write(base_env / ABI3_ARTIFACT, "build-host")`? That makes the assertion bite for any contents of the list, and it reads closer to what the fixture docstring already describes: a build host whose interpreter is not one the bundle targets.

phil-IO-p:
Regarding `scripts/build_addon.py:68` and  `:78-91` (not included in this diff)...

The Blender extension path may still be missing `awscrt`. `scripts/build_addon.py:68` pulls its requirement string out of the raw `pyproject.toml` text with `re.search(r"\"deadline\S* .+ .+\"", ...)`, which lands on the first quoted       
`deadline` line. Under this change that is the base `"deadline >= 0.60.4,< 0.61"` entry, which is exactly the one you deliberately kept the `console` extra off. That string then goes straight to `pip download --only-binary=:all: --   
python-version=3.11 --platform=...` at `:78-91`, so the extension's `wheels/` directory would have no awscrt in it.                                         
                                                                              
That path looks user-facing rather than dev-only: `README.md:39` points people at `releases/latest/download/index.json` as an extensions repository, and `.   github/actions/prepush_release_hook/action.yml:28` runs `hatch run build-addon` during the release workflow.                                           
                                                                              
Would it be worth injecting the extra there too? The extension targets 3.11 only, so it should get the `cp311-abi3` wheel with none of the per-version merge machinery, and reusing `depsBundle._add_console_extra` on that string   
would keep the two channels from drifting apart again. If the extension is knowingly out of scope for this fix, saying so in the PR description would save the next person the same trace.

scripts/depsBundle.py
```diff
@@ -12,8 +12,22 @@
...
+# _awscrt.cpython-3{9,10}-<platform>.so while 3.11+ get _awscrt.abi3.so. Resolving it only
+# in the base environment would ship whichever the build host produced, so any Blender whose
+# interpreter that single artifact does not cover would fail to import awscrt and AWS
+# Console sign-in would break there. (Blender 2.9's Python 3.9 is included here even though
+# requires-python's floor is the only place else that version appears: awscrt, pyyaml, and
+# xxhash all publish cp39 wheels for linux/mac/windows at the versions this bundle resolves,
+# verified via `pip download --no-deps --only-binary=:all: --python-version 3.9`.)
+# pyyaml is here because it ships a version-specific `_yaml` extension module: resolved only
+# in the base environment it lands built for a single interpreter, and pyyaml hides that by
+# falling back to its pure-Python parser on the others.
```
phil-IO-p: nit: Roughly 48% of the PR diff is comments; most of which represents the PR's own arguments being written into its source. Consider consolidating some of these

pyproject.toml
```diff
@@ -35,16 +35,28 @@ classifiers = [
...
+    # `console` extra at all. (BaseSubmitter/BaseSubmitterSettings, this package's other
+    # floor, first ship in 0.60.2, which 0.60.4 also satisfies.)
+    #
+    # The console extra is deliberately NOT requested here. It is only needed by the
+    # submitter, and it pulls awscrt, a compiled wheel. These base dependencies are
+    # resolved into the adaptor package by scripts/create_adaptor_packaging_artifact.sh
+    # with --only-binary=:all: --platform <tag>, and no awscrt wheel satisfying the
+    # 0.28.4 floor exists for the macosx_10_9_x86_64 tag that script uses, so console
+    # sign-in would appear installed and silently not work. The adaptor never takes the
+    # console path anyway: it runs on a worker with host-provided credentials.
```
phil-IO-p: I think the conclusion here is right but the stated reason is not, and the version number will drift. The `console` extra's only requirement is `botocore[crt]>=1.42.89`, and botocore's `crt` extra is an exact pin,         
`awscrt==0.36.0` — I could not find a 0.28.4 floor anywhere in that chain. The reason the `macosx_10_9_x86_64` tag cannot be satisfied looks more fundamental than a floor: awscrt publishes only `macosx_10_15_*` wheels, and I checked 0.27.6, 0.28.4 and 0.36.4, none of which has a `macosx_10_9_*` wheel at all.    
                                                                              
Would it be worth stating the durable half and dropping the number? Something like "awscrt ships no `macosx_10_9_*` wheel, so the extra cannot resolve under that tag" holds regardless of what botocore pins next, and nothing in this repo would notice the pin moving.

phil-IO-p (changes requested):

scripts/depsBundle.py
```diff
@@ -179,28 +196,57 @@ def _get_package_version(package: str, install_path: Path) -> str:
...
+    """Add deadline's `console` extra to a requirement string, preserving its specifier."""
+    match = re.fullmatch(
+        r"(?P<name>[A-Za-z0-9._-]+)(?:\[(?P<extras>[^\]]*)\])?(?P<spec>.*)", requirement
+    )
+    if not match or match.group("name").lower() != "deadline":
+        return requirement
+    extras = [extra for extra in (match.group("extras") or "").split(",") if extra]
+    if "console" not in extras:
+        extras.append("console")
+    return f"{match.group('name')}[{','.join(extras)}]{match.group('spec')}"
```
phil-IO-p: 1. The thing it guards against can still happen silently. It returns the requirement unchanged whenever the regex misses or the name is not exactly `deadline` lowercase, so a rename, a wrapped requirement, or `Deadline` would make the injection a no-op, pip would succeed, and the bundle would ship with 
no awscrt. `test_add_console_extra_changes_the_real_base_dependencies` catches that, but it lives in a suite the installer build does not run. Would a one-line check in `_build_base_environment` - that the rewrite changed at least one requirement, raising with what it looked for - be worth adding? That turns
the failure into a build error rather than a quiet shipping bug.              

2. More of a question than a request: `console` is now declared in two unrelated ways. `pyproject.toml` says `deadline[gui,console]` on the `gui` extra, but `_get_dependencies` only reads `project.dependencies`, so the bundle never sees that declaration and gets the extra purely from this rewrite. Neither enforces the other. Have you considered a dedicated extra that        
`_get_dependencies` reads, so the pyproject declaration is the single source and the build installs `.[bundle]` rather than reassembling the requirement string? If the rewrite stays, `packaging.requirements.Requirement` would replace the hand-rolled regex and give PEP 503 name normalization for free --  the test file already depends on `packaging`.

test/unit/deadline_submitter_for_blender/test_deps_bundle_native_merge.py
andychoquette: Correct — fixed in 1cf1cfb. Seeded the base env with a sentinel (`"build-host"`) instead of `supported_versions[-1]`, so the collision assertion cannot pass vacuously regardless of list shape; it now only passes if `_copy_native_to_base_env` actually overwrote the seeded value with the lowest-abi3-version tree's content.

scripts/depsBundle.py
andychoquette: Fair — trimmed in 1cf1cfb. Cut the comment blocks down to the durable constraints (abi3 non-uniformity, pyyaml's silent pure-Python fallback, the `--no-deps` double-direction rationale, the first-tree-wins/abi3-forward-compat rule in `_copy_native_to_base_env`, and the console-extra placement rationale), and removed the narrative/verification asides that were closer to PR argumentation than something a future reader needs from the source.

pyproject.toml
andychoquette: Confirmed, both parts. `pip download` for botocore>=1.42.89 shows `Requires-Dist: awscrt (==0.36.0) ; extra == "crt"` — an exact pin, not a floor. And checked 0.27.6, 0.28.4, and 0.36.4 directly against `macosx_10_9_x86_64` (`pip download --no-deps --only-binary=:all: --platform macosx_10_9_x86_64 awscrt==<version>`): none resolve. All of awscrt's mac wheels are `macosx_10_15_*`/universal2 or newer; `macosx_10_9_x86_64` has no matching wheel at any version I could find.

Reworded in 1cf1cfb to the durable claim (awscrt ships no `macosx_10_9_*` wheel) and dropped the version number.

scripts/depsBundle.py
andychoquette: Part 1 (build-time check): done in 1cf1cfb. `_build_base_environment` now raises if `_add_console_extra` left every dependency unchanged, naming what it looked for. This turns a rename/wrap/casing drift into a build failure instead of a silent ship-without-awscrt, and it runs on the actual installer build path (`build_deps_bundle` -> `_build_base_environment`), not just in the test suite.

Part 2 (dedicated extra vs. rewrite, and packaging.Requirement):

On the adaptor-leak concern — checked, and you're right that it does not actually apply to a dedicated extra. `_get_dependencies()` only reads `project.dependencies`; `create_adaptor_packaging_artifact.sh` only resolves that same list. Neither touches `project.optional-dependencies`. The existing `gui` extra already proves this: it nests `deadline[gui,console]` inside `project.optional-dependencies.gui` today and none of that leaks into the adaptor. So a dedicated `bundle` extra read by `_get_dependencies()` would not reintroduce the adaptor problem — that documented risk is specifically about `project.dependencies`, not extras in general.

That said, I'm keeping the rewrite for this PR rather than adopting either suggestion now:

- A dedicated extra changes `_build_base_environment`'s shape (it takes a `dependencies: list[str]` today, used identically by all 8 submitter repos in this fix wave; switching to `pip install .[bundle]` needs a package/sdist path instead) — a larger, cross-repo redesign that belongs with the "planned shared bundler" consolidation the code's docstrings already point at, not a one-off in blender while its siblings still use the rewrite.
- `packaging.Requirement` would give PEP 503 normalization for free, but `packaging` is currently declared only in requirements-testing.txt (the file's own comment says so) — it is not a build dependency of `depsBundle.py`, which runs via `depsBundle.sh` (`python3 scripts/depsBundle.py`) inside an internal CodeBuild project I can't inspect from this repo (`build_installers.yml` delegates to `aws-deadline/.github`'s reusable workflow, which hands off to an AWS CodeBuild project defined outside any repo I have access to). I can't confirm `packaging` is importable there, and I'd rather not gamble a Sev-2 fix on an unverified transitive dependency. For what it's worth, the "Deadline" casing example doesn't actually reproduce today — `_add_console_extra` already lowercases the matched name before comparing — so the regex's real gap is specifically rename/wrap/split, which part 1's build-time check now catches directly.

Happy to revisit both if the shared-bundler consolidation is scheduled soon — flagging it there would let all 8 repos move together.

andychoquette:
Confirmed the trace and fixed in 1cf1cfb.

`re.search(r"\"deadline\S* .+ .+\"", ...)` against the current pyproject.toml matches `"deadline >= 0.60.4,< 0.61"` (verified directly) — the base line, never the `gui` extra's `deadline[gui,console]` line further down. That requirement went straight into `pip download --platform=<tag>` with no console extra, so the extension's `wheels/` directory had no awscrt.

Injected the extra by reusing `depsBundle._add_console_extra` on the extracted string, exactly as suggested, via a new `_get_deadline_requirement()` in build_addon.py so the two channels (submitter bundle, extension wheels) can't drift apart again.

One thing your trace didn't flag that I hit while fixing it: adding the console extra pulls awscrt into the linux download, and awscrt publishes no `manylinux_2_28_x86_64`-tagged wheel (only `manylinux2014_x86_64`/`manylinux_2_17_x86_64`) — pip's tag matching for an explicit `--platform` does not walk up to a newer requested tag from an older wheel tag. `SUPPORTED_PLATFORMS`'s `manylinux_2_28_x86_64` entry would have made the linux download fail outright the moment the extra was added. Swapped it to `manylinux2014_x86_64`, verified via `pip download --only-binary=:all: --python-version=3.11 --platform=<tag> "deadline[console]>=<floor>"` against all four `SUPPORTED_PLATFORMS` entries (win_amd64, manylinux2014_x86_64, macosx_12_0_arm64, macosx_12_0_x86_64) — all resolve including awscrt now.

Also guarded the script's argparse/pipeline body behind `main()`/`if __name__ == "__main__":` so `_get_deadline_requirement()` is importable and unit-tested (`test/unit/deadline_submitter_for_blender/test_build_addon_console_extra.py`) without triggering a real build. The extension targets 3.11 only, confirmed no per-version merge machinery is needed there.

scripts/depsBundle.py
```diff
@@ -179,28 +189,60 @@ def _get_package_version(package: str, install_path: Path) -> str:
...
+
+
 def _build_base_environment(working_directory: Path, dependencies: list[str]) -> Path:
     (working_directory / "base_env").mkdir()
     base_env_path = working_directory / "base_env"
+    # Requested here rather than declared in project.dependencies: those also resolve into
+    # the adaptor package under a platform tag with no usable awscrt wheel (see
+    # pyproject.toml).
+    dependencies_for_pip = [_add_console_extra(dep) for dep in dependencies]
+    if dependencies_for_pip == dependencies:
```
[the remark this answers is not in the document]
andychoquette: Correct — fixed in 7485907. The guard now checks the postcondition (`_requests_console_extra`: some entry in `dependencies_for_pip` is a `deadline` requirement whose extras include `console`) rather than that the rewrite changed something. Verified both directions directly:

- `_build_base_environment(tmp_path, ["deadline[console]>=0.60.4,<0.61"])` — no longer raises (previously would have, since `_add_console_extra` is idempotent there and `dependencies_for_pip == dependencies`).
- `_build_base_environment(tmp_path, ["deadline-cloud>=0.60.4,<0.61"])` — still raises with a message naming what it looked for.

Added both as tests (`test_build_base_environment_accepts_a_deadline_requirement_already_requesting_console`, `test_build_base_environment_raises_when_nothing_requests_console`) and updated `test_add_console_extra_changes_the_real_base_dependencies` (renamed to `test_add_console_extra_makes_the_real_base_dependencies_request_console`) to pin the same postcondition against the real pyproject.toml dependencies. Factored the shared parsing out of `_add_console_extra` into `_parse_requirement` so `_requests_console_extra` isn't a second regex to keep in sync.

test/unit/deadline_submitter_for_blender/test_build_addon_console_extra.py
```diff
@@ -0,0 +1,61 @@
...
+]
+"""
+    assert _get_deadline_requirement(pyproject_contents) == "deadline[console] >= 0.60.4,< 0.61"
+
+
+def test_get_deadline_requirement_reads_the_base_line_not_the_gui_extra():
+    """Pins the pre-existing regex behavior this function wraps: it matches the FIRST quoted
+    `"deadline..."` line in the file, which is project.dependencies' unadorned entry -- not the
+    `gui` extra's `deadline[gui,console]` line further down. If pyproject.toml is ever
+    reordered so `gui` comes first, this documents the assumption in play, rather than letting
```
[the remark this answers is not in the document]
andychoquette: Correct on both counts — fixed in 7485907. Switched `_get_deadline_requirement` to parse `project.dependencies` via `tomllib` + `depsBundle._get_dependencies`/`_parse_requirement`, instead of `re.search` over the raw file text, so the extraction no longer depends on which `"deadline..."` line appears first.

Made the fixture adversarial as suggested (`gui` extra's `deadline[gui,console]` line placed before `dependencies` in the text) and verified it now genuinely exercises the property: with the old regex implementation this fixture returns `deadline[gui,console] >= 0.60.4,< 0.61` (fails the assertion); with the tomllib-based implementation it correctly returns `deadline[console]>=0.60.4,<0.61` regardless of textual order, since `tomllib.loads` builds the table dict independent of where each `[project...]` header appears in the file (confirmed both orderings parse to the same dict).

scripts/build_addon.py
```diff
@@ -12,24 +12,44 @@
...
     "win_amd64",
-    "manylinux_2_28_x86_64",
+    # awscrt (pulled in below via the console extra) publishes manylinux2014_x86_64 /
+    # manylinux_2_17_x86_64 wheels for linux, not manylinux_2_28_x86_64: `pip download
+    # --platform manylinux_2_28_x86_64` does not treat those as compatible (pip's tag
+    # generation for an explicit --platform only walks to older manylinux aliases, it does
+    # not walk up), so requesting the newer tag here would make this download fail outright
+    # once the console extra is requested. manylinux2014_x86_64 resolves every dependency
+    # this script downloads, including awscrt; verified via `pip download --only-binary=:all:
+    # --python-version=3.11 --platform=manylinux2014_x86_64 "deadline[console]>=<floor>"`.
```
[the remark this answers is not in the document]
andychoquette: Correct — the backtrack mechanism you describe (pip silently resolving an older, wheel-having version rather than failing) is exactly the macOS mechanism, and nothing on the linux channel caught the equivalent. Fixed in 7485907 with option 2: each `SUPPORTED_PLATFORMS` entry now downloads into its own destination (`_verify_platform_download`), which asserts a `deadline-*-py3-none-any.whl` at or above the requirement's `>=` floor is present, and that an `awscrt-*` wheel landed -- the actual payload this PR is about. Per-platform destinations (rather than the shared `{temp}/wheels`) so an earlier, compliant platform's deadline wheel can't mask a later platform's backtrack.

Also took the asymmetry note: reworded the `SUPPORTED_PLATFORMS` comment to state `manylinux2014_x86_64` resolved the closure "at the time of writing" / is a "point-in-time check... not a guarantee," and points at `_verify_platform_download` as what catches drift.

Added tests directly on `_verify_platform_download` (missing deadline wheel, deadline below floor, missing awscrt wheel, and the passing case) and on `_requirement_floor`.

test/unit/deadline_submitter_for_blender/test_console_signin_dependencies.py
```diff
@@ -0,0 +1,197 @@
...
+    assert deadline_reqs, "the gui extra declares no requirement on deadline"
+    for req in deadline_reqs:
+        assert "console" in req.extras, f"missing console extra in: {req}"
+
+
+def test_base_dependencies_do_not_request_the_console_extra(base_dependencies):
+    """Keeps awscrt out of the adaptor package.
+
+    The base list is resolved into the adaptor artifact under --only-binary=:all:
+    --platform <tag>. For macosx_10_9_x86_64 no awscrt wheel meets the floor, so pip
```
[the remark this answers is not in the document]
andychoquette: Correct — fixed in 7485907. Aligned both docstrings (module docstring and `test_base_dependencies_do_not_request_the_console_extra`) with pyproject.toml's corrected wording: awscrt ships no `macosx_10_9_*` wheel at all, so resolving the console extra there has no candidate to fall back to and the adaptor build fails outright, rather than "no wheel meets the floor, so pip resolves backwards." The constraint and the test are unchanged, only the stated mechanism.

scripts/build_addon.py
```diff
@@ -51,81 +121,96 @@ def strip_pyside6_wheel(whl_path: str) -> None:
...
+
+        # Download the wheels of the deadline library and its dependencies. Each platform
+        # gets its own destination -- not the shared {temp}/wheels -- so
+        # _verify_platform_download can check what THIS platform's resolution actually
+        # produced; a shared destination would let an earlier, compliant platform's deadline
+        # wheel mask a later platform's silent backtrack below the floor.
+        wheels_dir = Path(f"{temp}/wheels")
+        wheels_dir.mkdir(parents=True, exist_ok=True)
+        for platform in SUPPORTED_PLATFORMS:
+            platform_dest = f"{temp}/wheels_by_platform/{platform}"
```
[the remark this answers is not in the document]
andychoquette: Confirmed — `platform_dest` was nested under `temp`, and `temp` gets zipped wholesale into the shipped archive; the staged per-platform copies were also unstripped, so this was a real size regression invisible in `blender_manifest.toml`. Fixed in 686ddbf: the staging area now lives in its own `TemporaryDirectory()`, never under `temp`. Extracted the download loop into `_download_wheels()` and added `test_download_wheels_leaves_only_wheels_under_temp`, which asserts `temp` contains nothing but `wheels/` once the download completes -- would have failed against the previous code (it would have seen `{"wheels", "wheels_by_platform"}`).

scripts/depsBundle.py
andychoquette: Follow-up: two later commits (7485907, 686ddbf) added a fair amount of new code in the fix-response mode that drew this nit -- re-applied the same bar to everything they added/changed in 7e67e5d. Trimmed verification narrative and review-response argumentation ("this PR exists to...", restated rationale from other threads) and collapsed multi-paragraph test docstrings to one-line summaries, while keeping the constraints code cant show (pip backtracking silently and exiting 0, why staging lives outside `temp`, abi3/first-tree-wins, point-in-time wheel-availability checks). Comments/docstrings only -- no behavior, assertion, or test-name changes; verified the diff is prose-only. Leaving this thread open for you to resolve.

phil-IO-p (approved):

seant-aws (approved):
I think looks good. Tests seem fine and similar changes coming from another repo gives me good confidence. Thanks guys !

